PR #19, opened 2020-08-17
solved

nanccyy:


nanccyy:
> Looks good! The only thing - Misha said Object.create is an old method

Yes, you are right. 
Thanks for your advice.

PR #22, opened 2020-08-17
solution

furude-rika:


src/robotFactory.js
```diff
@@ -1,17 +1,80 @@
...
+  this.coords = coords;
+  this.chipVersion = chipVersion;
 }
 
-function FlyingRobot() {
-  // implement
+BaseRobot.prototype.goForward = function(step = 1) {
+  this.coords.y += step;
+
+  return this;
```
oggyrock, line 13: is it necessary?

furude-rika: agree

src/robotFactory.js
```diff
@@ -1,17 +1,70 @@
...
+};
+
+BaseRobot.prototype.goLeft = function(step = 1) {
+  this.coords.x -= step;
+};
+
+BaseRobot.prototype.goRight = function(step = 1) {
+  this.coords.x += step;
+
+  return this;
```
vsm2k17, line 25: if you want to call your methods in a chain you need to return this in every method not just in one, if you don't then you can remove it

src/robotFactory.js
```diff
@@ -1,17 +1,70 @@
...
+  this.coords.z -= step;
+};
+
+function DeliveryDrone(
+  name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+  this.maxLoadWeight = maxLoadWeight;
+  this.currentLoad = currentLoad;
 }
 
+DeliveryDrone.prototype = FlyingRobot.prototype;
```
vsm2k17, line 56: use object.setPrototypeOf(), this can break your code, no need to rewrite your prototype

src/robotFactory.js
```diff
@@ -1,17 +1,70 @@
...
+
+function FlyingRobot(name, weight, coords, chipVersion) {
+  this.coords = {
+    z: 0,
+  };
 }
 
-function DeliveryDrone() {
-  // implement
+FlyingRobot.prototype = BaseRobot.prototype;
```
vsm2k17, line 40: setPrototypeOf

PR #23, opened 2020-08-17
solution

Andrii256:


src/robotFactory.js
```diff
@@ -1,17 +1,69 @@
...
+  },
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion},`
+      + ` Weight: ${this.weight}`;
+  },
+};
 
-function DeliveryDrone() {
-  // implement
+// flying robot
```
vsm2k17, line 29: your function called flyingRobot, what for you need this comment then?

PR #26, opened 2020-08-17
Solution

rmnkk:


src/robotFactory.js
```diff
@@ -1,17 +1,89 @@
...
 }
 
-function FlyingRobot() {
-  // implement
+BaseRobot.prototype.goForward = function(step = 1) {
+  if (step > 1) {
+    this.coords.y += step;
+
+    return this;
+  }
```
DmJSCourse, line 15: Considering the default value being 1, 1 > 1 is a redundant comparison.
Also, this is more like a setter, so return is probably not needed.

src/robotFactory.js
```diff
@@ -1,17 +1,89 @@
...
+    this.coords.x -= step;
+
+    return this;
+  }
+};
+
+BaseRobot.prototype.goRight = function(step = 1) {
+  if (step > 1) {
+    this.coords.x++;
+
```
DmJSCourse, line 37: ++ will only increase number by 1, while you need to increase by step value

src/robotFactory.js
```diff
@@ -1,17 +1,89 @@
...
+    return this;
+  }
+};
+
+BaseRobot.prototype.getInfo = function() {
+  return `Robot: ${this
+    .name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
DmJSCourse, line 44: While this works in linter, it is definitely not a best practice. I suggest splitting it into multiple lines and adding with + on each line to make it more readable.

src/robotFactory.js
rmnkk: 👌

src/robotFactory.js
rmnkk: 👌

src/robotFactory.js
rmnkk: 👌

PR #31, opened 2020-08-17
Solution

AnastasiiaBortnichuk:
for some reason I don't understand, when I shorten the methods in each function, inheritance stops working, so I leave a lot of code

### DIFF
--- a/src/robotFactory.js
+++ b/src/robotFactory.js
@@ -1,16 +1,66 @@
 'use strict';
 
-function BaseRobot() {
-  // implement
+function BaseRobot(name, weight, coords, chipVersion) {
+  this.name = name;
+  this.weight = weight;
+  this.chipVersion = chipVersion;
+  this.coords = coords;
 }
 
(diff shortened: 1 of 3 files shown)


src/robotFactory.js
```diff
@@ -1,17 +1,67 @@
...
+  maxLoadWeight, currentLoad) {
+  FlyingRobot.call(this, name, weight, coords, chipVersion);
+  this.maxLoadWeight = maxLoadWeight;
+  this.currentLoad = currentLoad;
 }
 
+Object.setPrototypeOf(DeliveryDrone.prototype, FlyingRobot.prototype);
+
+DeliveryDrone.prototype.hookLoad = function(cargo) {
+  if (cargo.weight < this.maxLoadWeight) {
```
Elena-brv, line 56: if the weight of cargo equal to maxLoadWeight this condition should also work

AnastasiiaBortnichuk: Thanks! changed

PR #37, opened 2020-08-17
solution

Totowa95:


src/robotFactory.js
```diff
@@ -1,17 +1,69 @@
...
+) {
+  FlyingRobot.call(this, name, weight, chipVersion, coords);
+  this.maxLoadWeight = maxLoadWeight;
+  this.currentLoad = currentLoad;
 }
 
+Object.setPrototypeOf(DeliveryDrone.prototype, FlyingRobot.prototype);
+
+DeliveryDrone.prototype.hookLoad = function(cargo) {
+  if (cargo.weight <= this.maxLoadWeight) {
```
KaterinaSavvova, line 58: По условию: 
"DeliveryDrone prototype has method hookLoad, which saved the load object to currentLoad of DeliveryDrone instances **if weight of cargo less than** maxLoadWeight of drone"

src/robotFactory.js
```diff
@@ -1,17 +1,69 @@
...
+  this.coords.z += step;
+};
+
+FlyingRobot.prototype.goDown = function(step = 1) {
+  this.coords.z -= step;
+};
+
+function DeliveryDrone(
+  name, weight, chipVersion, coords, maxLoadWeight, currentLoad
+) {
```
vkryvytskyy, line 49:
```suggestion
function DeliveryDrone(
  name, weight, chipVersion, coords,
  maxLoadWeight, currentLoad,
) {
```

PR #42, opened 2020-08-18
factory

ChrisZinch:


src/robotFactory.js
```diff
@@ -1,17 +1,72 @@
...
+};
 
-function DeliveryDrone() {
-  // implement
+function FlyingRobot(name, weight, coords, chipVersion) {
+  this.name = name;
+  this.weight = weight;
+  this.coords = coords;
+  this.coords.z = 0;
+  this.chipVersion = chipVersion;
```
vkryvytskyy, line 34: you should transmit this values to the BaseRobot(use [call](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) method)

src/robotFactory.js
```diff
@@ -1,17 +1,72 @@
...
+Object.setPrototypeOf(FlyingRobot.prototype, BaseRobot.prototype);
+
+function DeliveryDrone(name, weight, coords, chipVersion,
+  maxLoadWeight, currentLoad) {
+  this.name = name;
+  this.weight = weight;
+  this.coords = coords;
+  this.chipVersion = chipVersion;
+  this.maxLoadWeight = maxLoadWeight;
+  this.currentLoad = currentLoad;
```
vkryvytskyy, line 55: you should transmit this values to the FlyingRobot(use [call](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) method)

src/robotFactory.js
```diff
@@ -1,17 +1,66 @@
...
+  },
+  goDown(step = 1) {
+    this.coords.z -= step;
+  },
+};
+
+Object.setPrototypeOf(FlyingRobot.prototype, BaseRobot.prototype);
+
+function DeliveryDrone(name, weight, coords, chipVersion,
+  maxLoadWeight, currentLoad) {
```
vkryvytskyy, line 46:
```suggestion
function DeliveryDrone(
  name, weight, coords, chipVersion,
  maxLoadWeight, currentLoad,
) {
```

PR #172, opened 2021-01-03
solution

NazarDovhanchuk:


vsm2k17 (changes requested):
!

src/robotFactory.js
```diff
@@ -1,17 +1,68 @@
...
-  // implement
+FlyingRobot.prototype = BaseRobot.prototype;
+
+FlyingRobot.prototype.goUp = function(step = 1) {
+  this.coords.z += step;
+};
+
+FlyingRobot.prototype.goDown = function(step = 1) {
+  this.coords.z -= step;
+};
```
varseniuk, line 45:
```suggestion
FlyingRobot.prototype = {
  goUp(step = 1) {
    this.coords.z += step;
  },
  goDown(step = 1) {
    this.coords.z -= step;
  },
};

Object.setPrototypeOf(FlyingRobot.prototype, BaseRobot.prototype);
```

It's better to use `setPrototypeOf`. And in such a case your code becomes more readable and less complicated

varseniuk (approved):
Take into consideration my comments

PR #179, opened 2021-01-21
Solution

georgy-dzumenko:


src/robotFactory.js
```diff
@@ -1,15 +1,114 @@
...
+      this.name}, Chip version: ${
+      this.chipVersion}, Weight: ${
+      this.weight}`;
+  };
 }
 
-function FlyingRobot() {
-  // implement
+FlyingRobot.prototype = new BaseRobot();
+DeliveryDrone.prototype = new FlyingRobot();
```
vkryvytskyy, line 51: use setPrototypeOf to define prototype

vkryvytskyy (changes requested):

src/robotFactory.js
georgy-dzumenko: I used in the new version

vkryvytskyy (approved):

PR #265, opened 2021-05-12
solution

Namart:


src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, position, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = position.x ? position : {
+      x: 0, y: 0,
+    };
```
igor-stasiv, line 10: What if position.x = 0 or doesn't exist and position.y = 1? It will ignore the value of position.y and will set coords.x and coords.y equal 0

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, position, chipVersion) {
+    super(name, weight, position, chipVersion);
+
+    this.coords = position.z ? position : {
+      x: 0,
+      y: 0,
+      z: 0,
```
igor-stasiv, line 43: Here the same situation

igor-stasiv (approved):
I approve because it works, but pay attention to my suggestions

src/robotFactory.js
varseniuk: Igor is right. Let's check and use both (x and y) coordinates and pass them to `this.coords`

src/robotFactory.js
varseniuk: Yep

varseniuk (changes requested):
Great start! But let's handle all coordinates

src/robotFactory.js
```diff
@@ -1,10 +1,69 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, position, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
```
varseniuk, line 7:
```suggestion
    this.coords = {
      x: position.x ? position.x : 0,
      y: position.y ? position.y : 0,
    };
```
You may assign values directly here

varseniuk (approved):
Good! 👍 
Left one comment

PR #266, opened 2021-05-12
Solution

igor-stasiv:


provolovskyi-illia (approved):
Очень хорошая работа

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
```
varseniuk, line 7:
```suggestion
    this.coords = {
      x: coords.x ? coords.x : 0,
      y: coords.y ? coords.y : 0
    };
```
Why won't you assign these values right here?

varseniuk (approved):
Great! 🔥 
Left one comment

src/robotFactory.js
igor-stasiv: Totally agree with you, it will be really more obvious

PR #274, opened 2021-05-13
Solution

KapToH337:


src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null) {
+      this.currentLoad = cargo;
+    }
+
+    if (this.currentLoad.weight > this.maxLoadWeight) {
```
victor-arutiunov, line 72: You can combine this "if" with previous like "**_if (this.currentLoad === null && cargo.weight < this.maxLoadWeight)_**"

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    if (!coords.x) {
+      coords.x = 0;
+      coords.y = 0;
```
mykola-yarchuk, line 7: don't change the coords objects from arguments, create new variables and use them

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+  };
+
+  goLeft(a = 1) {
+    this.coords.x -= a;
+  };
+
+  getInfo() {
+    const n = this.name;
+    const c = this.chipVersion;
+    const w = this.weight;
```
mykola-yarchuk, line 35: don't create new variables just use string concatenation

src/robotFactory.js
mykola-yarchuk: great advice from Victor

mykola-yarchuk (changes requested):

src/robotFactory.js
mykola-yarchuk: here you assign 0 to object `coords` from function's parameter, that is not okay, create new variable
```suggestion
      const { x = 0, y = 0 } = coords;
```

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    if (!coords.x) {
+      coords.x = 0;
+      coords.y = 0;
+    }
 
-class FlyingRobot {}
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
```
mykola-yarchuk, line 12: and use x and y here

mykola-yarchuk: or
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    }
```

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):

PR #283, opened 2021-05-13
Solution

avgusta92:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion = 0
```
YuriiLiso, line 8:
```suggestion
    chipVersion
```
Isn't necessary.

YuriiLiso (approved):
Looks Great!

varseniuk (approved):
Good 👍

src/robotFactory.js
varseniuk: Agree

PR #287, opened 2021-05-13
created robot classes

YuriiLiso:


Bogdan-Kotsupey (approved):
nice

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
 
-class FlyingRobot {}
+    if (Object.keys(coords).length === 0) {
+      this.coords = {
+        x: 0,
+        y: 0,
+      };
+    } else {
+      this.coords = coords;
+    }
```
varseniuk, line 16:
```suggestion
    this.coords = {
      x: coords.x ? coords.x : 0,
      y: coords.y ? coords.y : 0
    };
```

You may use smth like this to make it shorter and to avoid redundant checks

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+
+    if (Object.keys(coords).length < 3) {
+      this.coords = {
+        x: 0,
+        y: 0,
+        z: 0,
+      };
+    } else {
+      this.coords = coords;
+    }
```
varseniuk, line 54: Same here

varseniuk (approved):
Nice work!
Left a comment on how to make it more simple

PR #299, opened 2021-05-14
solution

Allrightman:


src/robotFactory.js
```diff
@@ -1,10 +1,130 @@
...
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: `
+      + `${this.chipVersion}, Weight: ${this.weight}`;
+  }
```
den-bulaev, line 36: this methods should be implement in the BaseRobot class

src/robotFactory.js
```diff
@@ -1,10 +1,130 @@
...
+    super.goRight();
+  }
+
+  goLeft(step = 1) {
+    super.goLeft();
+  }
+
+  getInfo() {
+    super.getInfo();
+  }
```
den-bulaev, line 126: you repeat yourself

src/robotFactory.js
```diff
@@ -1,10 +1,130 @@
...
```
den-bulaev, line 77: you can write this methods only once in parent class(BaseRobot) and use them in child classes (thanks to inheritance)

den-bulaev (approved):
Good job

src/robotFactory.js
Allrightman: So they are (BaseRobot = line 3 to line 37)

src/robotFactory.js
Allrightman: Oh, that's right! 
Overcoding)))

VinnieJ-2k20 (approved):
Good job 👍

PR #301, opened 2021-05-16
solution

Nazar-Kudatsky:


leonid-vegera (approved):
Great work!

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad
```
VinnieJ-2k20, line 57: Later you check if `this.currentLoad === null`. Let's set a default value not to have `undefined` here.

```suggestion
    currentLoad = null
```

VinnieJ-2k20 (approved):
Good job 👍 

Left one comment.

PR #306, opened 2021-05-17
solution

IvanKolodiazhnyi:


src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+      y,
+      z,
+    };
+  };
+
+  goUp(step = 1) {
+    this.coords.z += step;
+
+    return this;
+  };
```
aliceblunt8, line 65: no need to return this in this task

aliceblunt8 (approved):

PR #307, opened 2021-05-17
Solution 1

Grygoriy-Shytikov:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+  };
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`; //eslint-disable-line
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    const { x = 0, y = 0, z = 0 } = coords;
```
VinnieJ-2k20, line 40: No need to reassign `x` and `y` here, it is done in BaseRobot.

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
VinnieJ-2k20, line 59: Later you check if `this.currentLoad === null`. It is better to set it as default to avoid potential problems with `undefined`.
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null,
  ) {
```

VinnieJ-2k20 (changes requested):
Almost done.
Let's make some final changes.

VinnieJ-2k20 (approved):
Good job 👍

PR #310, opened 2021-05-17
Solution

starzhitskiy:


src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
...
+  goUp(step = 1) {
+    this.coords.z += step;
+  }
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
VinnieJ-2k20, line 51: Later you check if `this.currentLoad === null`. It is better to add it by default to avoid potential problems with `undefined`.
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null,
  ) {
```

VinnieJ-2k20 (approved):
Good job 👍 

Added one comment.

PR #312, opened 2021-05-18
solution

dmitry-puhliakov:


src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+      + `Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = coords;
```
VinnieJ-2k20, line 44: No need to reassign `x` and `y` here, it is done in `BaseRobot`.

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
VinnieJ-2k20, line 63: Later you check if `this.currentLoad === null`. It is better to set it by default to avoid potential problems with `undefined.`
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null,
  ) {
```

VinnieJ-2k20 (changes requested):
Nice solution.
Need couple more small fixes.

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+      + `Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords.z = coords.z ? coords.z : 0;
```
VinnieJ-2k20, line 44:
```suggestion
    this.coords.z = coords.z || 0;
```

VinnieJ-2k20 (approved):

PR #380, opened 2021-08-12
robot factory made

ihor-jpeg:


src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
```
Anna-Tkachenko, line 9: 👍

Anna-Tkachenko (approved):
Well Done🔥

src/robotFactory.js
VinnieJ-2k20, line 9: It is better to fill object in place rather than by separate props assignments.
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment on code style.

PR #381, opened 2021-08-12
Solution

Anna-Tkachenko:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+    return `Robot: ${this.name}, `
+      + `Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z ? coords.z : 0;
```
ihor-jpeg, line 41: you can use a logical operator here, to make the assignment. I recently found out about their amazing utility in assigning values to the variables when there is a condition

ihor-jpeg (approved):
nicely done! 👍

VinnieJ-2k20 (approved):
Good job 👍

PR #388, opened 2021-08-15
solution

SashaRudenko:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = coords.x === undefined || coords.y === undefined
+      ? {
+        x: 0,
+        y: 0,
+      }
+      : coords;
+  }
```
SashaLifashkin, line 15: Так тоже работает! Но можно проще записать, через логическое ИЛИ.
this.coords = coords.x || 0;
Кода будет меньше.

SashaLifashkin:
Зачёт! )

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords.z = coords.z === undefined
+      ? 0
+      : coords.z;
```
VinnieJ-2k20, line 45: Same here. Use short conditional assignment.

VinnieJ-2k20 (changes requested):
Almost done, just need to fix coords assignment

src/robotFactory.js
VinnieJ-2k20: Condition is incorrect. If you'll have valid object with `{ x: 25 }`, `y` will be `undefined` and `x` will write as `0`.
Sasha suggested correct solution in the comments.

VinnieJ-2k20 (approved):
Good job 👍

PR #393, opened 2021-08-18
Sol

tumann21:


src/robotFactory.js
```diff
@@ -1,10 +1,90 @@
...
+    }
+  }
+
+  unhookLoad() {
+    this.currentLoad = null;
+  }
+}
+
+Object.setPrototypeOf(FlyingRobot.prototype, BaseRobot.prototype);
+Object.setPrototypeOf(DeliveryDrone.prototype, FlyingRobot.prototype);
```
vkryvytskyy, line 87: use `extends` keyword instead of setPrototypeOf

src/robotFactory.js
```diff
@@ -1,10 +1,90 @@
...
+
+class DeliveryDrone {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
+    this.chipVersion = chipVersion;
+  }
```
vkryvytskyy, line 73: use `super` keyword here to call parents constructor

vkryvytskyy (changes requested):

PR #409, opened 2021-08-20
Robot factory Solution

TarasHoliuk:


src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
```
illiarerun, line 4: FYI: Default value for coords would work only if consstructor was called with only two first params (name and weight), also chipVersion param wouldn't have any default value. I suggest removing default value cause we always call this constructor with atleast blank object for coords.
```suggestion
  constructor(name, weight, coords, chipVersion) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
...
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+     + `Chip version: ${this.chipVersion}, `
+     + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
```
illiarerun, line 43: Same here with default value
```suggestion
  constructor(name, weight, coords, chipVersion) {
```

illiarerun (approved):
Good job 🔥 But check my comments about default values 🙏

src/robotFactory.js
TarasHoliuk: I removed default value, thank you for advice.

src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
...
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null
+      && cargo.weight <= this.maxLoadWeight
+    ) {
```
varseniuk, line 76:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight
    ) {
```

varseniuk (approved):
👍

PR #410, opened 2021-08-20
first try

zadorojniiarsen:


src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+    weight,
+    coords,
+    chipVersion
+  ) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    if (Object.keys(coords).length === 0) {
```
TarasHoliuk, line 14: You can use destructuring to declare coords values

TarasHoliuk (approved):
Left one comment, everything else is good

src/robotFactory.js
VinnieJ-2k20: This check is incorrect. Find another way to check existense of `x` and `y` props in `coords`

src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+
+    if (coords.z === undefined) {
+      this.coords = {
+        x: 0,
+        y: 0,
+        z: 0,
+      };
+    } else {
+      this.coords = coords;
+    }
```
VinnieJ-2k20, line 59: Apply only `z` coordinate here.

src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null) {
+      if (cargo.weight <= this.maxLoadWeight) {
```
VinnieJ-2k20, line 81: Don't write nested `if`s. Use logical operator in condition.

VinnieJ-2k20 (changes requested):
Solution works, but needs some fixes. Details in comments.

VinnieJ-2k20 (approved):

PR #422, opened 2021-08-22
task is done

Oleksandr-Zhitenko:


src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+}
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad == null && cargo.weight <= this.maxLoadWeight) {
```
VinnieJ-2k20, line 60: You should never use `==` comparison
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

VinnieJ-2k20 (changes requested):
Almost done, just fix one small thing.

varseniuk (approved):

PR #451, opened 2021-09-30
js_robot_factory done

AndreyMalyshev147:


oksana-onopa (approved):
LGTM 👍

src/robotFactory.js
```diff
@@ -1,10 +1,122 @@
...
+    super.goBack();
+  };
+
+  goRight() {
+    super.goRight();
+  };
+
+  goLeft() {
+    super.goLeft();
+  };
```
Anna-Tkachenko, line 67: It's redundant here

src/robotFactory.js
```diff
@@ -1,10 +1,122 @@
...
+
+  goUp() {
+    super.goUp();
+  };
+
+  goDown() {
+    super.goDown();
+  };
```
Anna-Tkachenko, line 108: Same here

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!👍

PR #523, opened 2021-11-17
Solution

dariia-romanova:


SerhiiKirik (approved):
Very nice, such pure code 🔥

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return 'Robot: ' + this.name
+    + ', Chip version: ' + this.chipVersion
+    + ', Weight: ' + this.weight;
```
ihor-jpeg, line 35: I think interpolation here is the way to go

ihor-jpeg (approved):
Good stuff!

src/robotFactory.js
dariia-romanova: Used concatenation, because the line was too long. Is it better to use both like this?
```suggestion
    return `Robot: ${this.name}, `
      + `Chip version: ${this.chipVersion}, `
      + `Weight: ${this.weight}`;
```

PR #531, opened 2021-11-23
Solution js_robot_factory

anastasiiaafanasieva:
Solution js_robot_factory

### DIFF
--- a/src/robotFactory.js
+++ b/src/robotFactory.js
@@ -1,10 +1,84 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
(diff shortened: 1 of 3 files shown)


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    if (coords.x === undefined) {
```
vkryvytskyy, line 8:
```suggestion
    if (!coords.x) {
```

vkryvytskyy (approved):

PR #540, opened 2021-11-25
solution

4qwerty:


WorkAccCom (approved):
Good!

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && !this.currentLoad) {
+      this.currentLoad = cargo;
+    }
+  }
```
ihor-jpeg, line 64: the task does not mention this case, but you can think what happens when we first hook a load with weight 0; it will behave like we cannot hook a new load, because of this line `!this.currentLoad`. as 0 will equal to true in this case. not a mistake, just something ti think about

ihor-jpeg (approved):
Noiiice

PR #541, opened 2021-11-25
Solution to robot factory

Vitalii-Lytvyn01:


src/robotFactory.js
```diff
@@ -1,10 +1,63 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
```
MagicBox15, line 7:
```suggestion
```
why did you separate this property? if you have next two lines, you don't need the first one

src/robotFactory.js
```diff
@@ -1,10 +1,63 @@
...
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${
+      this.chipVersion}, Weight: ${this.weight}`;
```
MagicBox15, line 27: I think it's better to use concat here. Cause this part looks not very nice
```suggestion
     return `Robot: ${this.name}, `
    + `Chip version: ${this.chipVersion}, `
    + `Weight: ${this.weight}`;
```

Vitalii-Lytvyn01: Yeah, it was mostly because of linter error about too many chars in one line.

src/robotFactory.js
illiarerun: Agree

illiarerun (changes requested):
Good job 🔥 Fix issue that Vlad commented about:

src/robotFactory.js
```diff
@@ -1,10 +1,64 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
```
illiarerun, line 9: You can shorten this three assignments:
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

illiarerun (approved):
Good job 🔥  Make sure to check this one comment:

PR #543, opened 2021-11-25
Solution of robot_factory

Dima-Ziakhor:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.8.10",
```
WorkAccCom, line 19: Please do not add useless files to your PR.
You can use **git add src/** command to add needed files
**git status** command — for checking file status

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(value = 1) {
```
WorkAccCom, line 15: maybe a "step" is a much more appropriate name to this parameter than the "value"

WorkAccCom (approved):

ihor-jpeg (approved):
well done. Let work on those comments below

PR #545, opened 2021-11-25
Solution

Veronika-Korona:


lanebx (approved):
GJ)

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: 0,
+      y: 0,
+      z: 0,
+      ...coords,
+    };
```
illiarerun, line 48: You already set coords in parent consturctor, so you only need to set z coord here

illiarerun (changes requested):
Good job 🔥  Let's fix one little issue:

aionova (approved):
GJ!

PR #554, opened 2021-11-26
First solution

WorkAccCom:


tmrsl (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
 
-class DeliveryDrone {}
+  getInfo() {
+    return `Robot: ${
+      this.name
+    }, Chip version: ${
+      this.chipVersion
+    }, Weight: ${
+      this.weight
+    }`;
```
ihor-jpeg, line 22: this is not easy to read

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      z: coords.z || 0,
+      x: coords.x || 0,
+      y: coords.y || 0,
```
ihor-jpeg, line 49: but you already do this in the BaseRobot constructor.

ihor-jpeg (changes requested):
Good work! Just a few minor details to have a look at

aionova (approved):

PR #608, opened 2022-01-03
sm msg

Mikedacheka:


Valeria-Valuyskaya (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    x: 0,
+    y: 0,
+  }, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
Denchepornyuk, line 13: checking with || is unnecessary here, you set a default value for `x` and `y` in the function arguments
x and y always will be number here

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = {
+    z: 0,
+  }, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      ...coords,
+      z: coords.z || 0,
```
Denchepornyuk, line 48: the same situation here

Denchepornyuk (changes requested):

src/robotFactory.js
Mikedacheka: Why it doesn't work with default params?Tests with z-param didn't pass
Got some ideas? 
Can we set as default object props?

Denchepornyuk (approved):

PR #612, opened 2022-01-04
Solution

AlexAzalor:


src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this;
```
Mikedacheka, line 18: don't need to return this;

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+  }
+
+  goBack(step = 1) {
+    this.coords.y -= step;
+
+    return this;
```
Mikedacheka, line 24: same with all this in methods

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
+
+    this.coords = {
```
Mikedacheka, line 52: you have x,y from BaseRobot

Mikedacheka (changes requested):
some minor fixes

Denchepornyuk (approved):

PR #616, opened 2022-01-04
save

bcmp4p1:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
gnosis-frog, line 44: the parent class already has this x, y coords

gnosis-frog (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+      z: coords.z || 0,
+    };
```
Elena-brv, line 46: add only z coordinates

Elena-brv (changes requested):

Elena-brv (approved):

PR #645, opened 2022-02-04
Robot factory

polosanya:


Vadym-Mishchenko:
Well Done! 👍

Vadym-Mishchenko (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    this.currentLoad = this.currentLoad === null
+      && cargo.weight <= this.maxLoadWeight
+      ? cargo
+      : this.currentLoad;
```
Anna-Tkachenko, line 67: It's better to use if/else statement here - will be more understandable

Anna-Tkachenko (approved):

PR #664, opened 2022-02-06
Solution

oleksandr-kozhushko:


src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
Anna-Tkachenko, line 45: Flying robot should add only z coords, x and y are assigned in Base robot

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #690, opened 2022-03-03
Solution for js_robot_factory

murat-yesimbietov:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 } = {}, chipVersion) {
```
Bohdan-Maksymiuk, line 4: ??
```suggestion
  constructor(name, weight, coords, chipVersion) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x,
+      y,
```
Bohdan-Maksymiuk, line 10:
```suggestion
      x: coords.x || 0,
      y: coords.y || 0,
```

Bohdan-Maksymiuk (changes requested):
Good job! Rewrite instances as I show. It is better for readability and code maintaining

PR #695, opened 2022-03-13
solution 

djtyrf312:


src/robotFactory.js
```diff
@@ -1,10 +1,63 @@
...
+    this.coords.y -= step;
+  };
+  goRight(step = 1) {
+    this.coords.x += step;
+  };
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
Bohdan-Maksymiuk, line 27:
```suggestion
        return `Robot: ${this.name}`
      + `, Chip version: ${this.chipVersion}`
      +`, Weight: ${this.weight}`;
```

Bohdan-Maksymiuk (changes requested):
Check tests before pushing your solution

src/robotFactory.js
djtyrf312: sorry

PR #723, opened 2022-04-15
Solution

Anastasiia-Svintsova:


src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    if (this.coords.z === undefined) {
+      this.coords.z = 0;
+    }
```
ivan-kadykalo, line 44: It works fine, but in the checklist says -> 

2. [CODE KNOWLEDGE] - use logic operator `||` if you want safely add some initial values.

you can write `this.coords.z = coords.z || 0` and it will work the same

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+      coords.x = 0;
+      coords.y = 0;
+    }
 
-class FlyingRobot {}
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
+    this.chipVersion = chipVersion;
+  }
```
ivan-kadykalo, line 14: 2. [CODE KNOWLEDGE] - use logic operator `||` if you want safely add some initial values.

Here you can avoid this if statement
```
this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

ivan-kadykalo (changes requested):
Good job, two things left

PR #737, opened 2022-04-19
add task

Helexi:


src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(str) {
```
RuslanaTomnyuk, line 69: has hookLoad method taking a cargo object!

RuslanaTomnyuk (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+    if (this.currentLoad === null && str.weight <= this.maxLoadWeight) {
```
MaksimusCRS, line 70:
```suggestion
    if (!this.currentLoad && str.weight <= this.maxLoadWeight) {
```

src/robotFactory.js
MaksimusCRS, line 69: why str here?

MaksimusCRS (changes requested):

MaksimusCRS (approved):

PR #739, opened 2022-04-19
solved

misha200119:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+  };
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.currentLoad = null;
+    this.maxLoadWeight = maxLoadWeight;
+    this.hookLoad(currentLoad);
+  }
```
YegorKochetkov, line 62:
```suggestion
  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
    super(name, weight, coords, chipVersion);
    this.currentLoad = currentLoad || null;
    this.maxLoadWeight = maxLoadWeight;
  }
```

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+
+  hookLoad(cargo) {
+    if (cargo !== null && cargo !== undefined
+          && this.currentLoad === null
+          && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    }
+  }
```
YegorKochetkov, line 70:
```suggestion
  hookLoad(cargo = 0) {
    if (cargo.weight <= this.maxLoadWeight) {
      this.currentLoad = this.currentLoad || cargo;
    }
  }
```

YegorKochetkov (approved):
in principle, everything is correct. but it can be written more briefly

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords
+    = coords.hasOwnProperty('x') && coords.hasOwnProperty('y')
+        ? coords
+        : {
+          x: 0,
+          y: 0,
+        };
```
MaksimusCRS, line 15:
```suggestion
    constructor(name, weight, coords, chipVersion) {
   
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

src/robotFactory.js
MaksimusCRS: Fix this also

src/robotFactory.js
MaksimusCRS: And this

MaksimusCRS (changes requested):

src/robotFactory.js
misha200119: if i try get prop from cargo before check to null i take exception Cannot read property 'weight' of null so...

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+    maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+    this.currentLoad = null;
+    this.maxLoadWeight = maxLoadWeight;
+    this.hookLoad(currentLoad);
+  }
+
+  hookLoad(cargo) {
+    if (cargo !== null
+      && cargo.weight <= this.maxLoadWeight) {
```
MaksimusCRS, line 61:
```suggestion
    if (!cargo && cargo.weight <= this.maxLoadWeight) {
```

MaksimusCRS (approved):

PR #778, opened 2022-05-16
solution

vlad-demchuk:


alexandra-protyanova (approved):
Amazing!

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+    this.coords = {};
 
-class FlyingRobot {}
+    if (coords.y && coords.x) {
+      this.coords = coords;
+    } else {
+      this.coords = {
+        x: 0,
+        y: 0,
+      };
```
Anna-Tkachenko, line 15: Check point 2 in checklist and use simpler way to initialize coords props
this.coords = {
  x: coords.x || 0,
  ...
}

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    if (coords.z) {
+      this.coords.z = coords.z;
+    } else {
+      this.coords.z = 0;
+    }
```
Anna-Tkachenko, line 51: Same here, use simpler way to initialize z using || operator

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #788, opened 2022-05-17
Robot factory was done

vlasiuk-anatolii:


src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x,
+      y: coords.y,
```
vlad-demchuk, line 45: i gues, you already have x and y, it's not neccesary to set them again

vlad-demchuk (approved):

Anna-Tkachenko (approved):
⭐️

PR #789, opened 2022-05-17
solution

AnastasiyaSopotnytska:


src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, `
+    + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot {
```
dellachaise, line 37: Why you didn't use inheritance like in line 59?

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+    }
+  }
+
+  unhookLoad() {
+    this.currentLoad = null;
+  }
+}
+
+Object.setPrototypeOf(FlyingRobot.prototype, BaseRobot.prototype);
+Object.setPrototypeOf(DeliveryDrone.prototype, FlyingRobot.prototype);
```
dellachaise, line 78: I think this line is redundant because you already add inheritance for tis object in line 59, isn't you?

dellachaise (approved):
Great job!

Please look at comments.

Anna-Tkachenko (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.name = name;
+    this.weight = weight;
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
Anna-Tkachenko, line 47: x and y properties you set in BaseRobot constructor
delete them from FlyingRobot constructor

Anna-Tkachenko: same with name, weight and chipVersion props

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #792, opened 2022-05-17
Solution

JustDan17:


src/robotFactory.js
```diff
@@ -1,10 +1,107 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords = {
+      x: 0, y: 0, z: 0,
```
ZUBOV-ILLIA, line 8: For base robot not need `z` coordinates.
Also in 17 and 18 lines you choosing the value, so here its not necesary

src/robotFactory.js
```diff
@@ -1,10 +1,107 @@
...
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this;
```
ZUBOV-ILLIA, line 25: not necesary return something here.
Also lines 31, 37, 43

ZUBOV-ILLIA (changes requested):
Good job!
Just small changes.

Anna-Tkachenko (approved):
⭐️

PR #796, opened 2022-05-17
done

uliton:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
```
Anna-Tkachenko, line 45: x and y properties was set in BaseRobot constructor, remove them from Flying robot constructor

Anna-Tkachenko (changes requested):

src/robotFactory.js
uliton: I don't know why, but it doesn't work when i remove it!

Anna-Tkachenko (approved):
⭐️

PR #797, opened 2022-05-17
solution

lytvyshko:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
 
-class FlyingRobot {}
+    if (coords.x === undefined) {
+      this.coords = {
+        x: 0,
+        y: 0,
+      };
+    } else {
+      this.coords = coords;
+    }
```
Anna-Tkachenko, line 16: Check point 2 in checklist

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    if (coords.z === undefined) {
+      this.coords.z = 0;
+    }
```
Anna-Tkachenko, line 47: Same here

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #807, opened 2022-05-18
Solution

huberciak:


andreisucman (approved):
Good job

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
meforoma, line 55: better multi-line properties when there are more than 4 of them:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad
  ) {
```

meforoma (approved):
good one 👍 
please consider the only code-style comment for the future

PR #810, opened 2022-05-18
First_commit

igoree88:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
Anna-Tkachenko, line 45: You should add only z property to this.coords object here

Anna-Tkachenko (approved):

PR #814, opened 2022-05-19
Solution

PRO-GRAM-MER:


rHashTag (approved):
LGTM!

One comment on the code, I can see some unnecessary files were committed. Try not to use `git commit -am` but rather add the files then commit with `git commit -m`

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
 'use strict';
 
-class BaseRobot {}
+// class BaseRobot {}
```
meforoma, line 3: no need to leave commented code
please remove all commented code lines

meforoma (approved):
good job 👍 
but don't forget to cleanup commented code

PR #816, opened 2022-05-20
solution

Sardorbek1122:


amitpatiljc (approved):
Good job 👍

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name,
```
meforoma, line 52: good that you multi-line props, but you need to start first prop at new line too))
```suggestion
  constructor(
    name,
```

meforoma (approved):
good one @Sardorbek1122 
one tiny comment - when multi-lining props, start the first one from new line too - see comment in the code

PR #818, opened 2022-05-22
solution

WojciechCzarnowski:


amitpatiljc (approved):
Good job 👍

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
meforoma, line 53: usually it's better to multi-line props when there are more than 4:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad
  ) {
```

meforoma (approved):
good job @WojciechCzarnowski 
just take a look at one code-style comment

PR #840, opened 2022-06-18
Solution

ik-web:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, cargo) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = cargo;
```
jstmpelowycz, line 59: down below, you check `currentLoad` for being null, so I guess you need this
```suggestion
    this.currentLoad = cargo || null;
```

jstmpelowycz (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    const { name: naming, chipVersion: chipVers, weight: dimensions } = this;
```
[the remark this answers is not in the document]
ik-web: Fixed

src/robotFactory.js
ik-web: Fixed

jstmpelowycz (approved):

PR #842, opened 2022-06-19
add solution

Galyopa:


src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = null || currentLoad;
```
jstmpelowycz, line 76: why not to make it simpler?
```suggestion
    this.currentLoad = currentLoad || null;
```

jstmpelowycz (approved):

PR #847, opened 2022-06-21
Solution

ostapfriend:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && this.currentLoad === null) {
+      this.currentLoad = cargo;
```
AndriiNychaiuk, line 68: Great job! The only (but not necessary) thing is that check for null or undefined you can make like this 
```if (!this.currentLoad) {}```

**Review by AndriiNychaiuk — APPROVED** [review]

File: src/robotFactory.js [hunk 2]
```diff
@@ -1,10 +1,78 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
jstmpelowycz, line 44: you don't need to override these properties, they're already there

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
```
jstmpelowycz, line 63: you'd rather make sure that `undefined` is not written
```suggestion
    this.currentLoad = currentLoad || null;
```

jstmpelowycz (changes requested):

src/robotFactory.js
illiarerun: Not only for `undefined`, but for `0` too, fix it

illiarerun (changes requested):

jstmpelowycz (approved):

PR #859, opened 2022-06-22
solution

Victance:


src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: `
+    + `${this.chipVersion}, Weight: ${this.weight}`;
```
LesyaSvk, line 35: You can also use here destructuring, how you did it with coords)

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    const { x = 0, y = 0, z = 0 } = coords;
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+
+    this.coords = {
+      x: x,
+      y: y,
+      z: z,
+    };
```
LesyaSvk, line 74: Do you really need this code block ? DeliveryDrone has already inherited this property from FlyingRobot.

LesyaSvk (changes requested):

src/robotFactory.js
Victance: getInfo() {

    return `Robot: ${name}, Chip version: `
    + `${chipVersion}, Weight: ${weight}`;
  }
  
  Like this?
  I think it is not worth it since we use this destruсturing only once

mykola-koval (approved):

PR #861, opened 2022-06-23
Solution

Tsymbal-Viacheslav:


illavodonis (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,97 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
jstmpelowycz, line 4: what if `coords` is undefined?

src/robotFactory.js
```diff
@@ -1,10 +1,97 @@
...
+  };
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
jstmpelowycz, line 53: you don't need it

src/robotFactory.js
```diff
@@ -1,10 +1,97 @@
...
+    return this;
+  };
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
```
jstmpelowycz, line 76: better add 
```
```
thus, if `currentLoad` is not passed, `undefined` won't be written

src/robotFactory.js
```diff
@@ -1,10 +1,97 @@
...
+
+  hookLoad(cargo) {
+    if (this.maxLoadWeight < cargo.weight || (this.currentLoad)) {
+      return this;
+    }
+
+    this.currentLoad = cargo;
+
+    return this;
+  };
```
jstmpelowycz, line 87: do you need this function to return `this`?

src/robotFactory.js
```diff
@@ -1,10 +1,97 @@
...
+
+  unhookLoad() {
+    this.currentLoad = null;
+
+    return this;
```
jstmpelowycz, line 92: the same way here

jstmpelowycz (changes requested):

illiarerun (approved):
GJ

PR #864, opened 2022-06-23
Solution

yanaost:


src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
Samaev, line 44: You don`t have to put X and Y again, they are already extended from BaseRobot

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
+  }, chipVersion, maxLoadWeight, currentLoad) {
```
Samaev, line 61:
```suggestion
  constructor(name, weight, coords , chipVersion, maxLoadWeight, currentLoad) {
```

that is an extra text, just link 'coords' would be enough and workable. You already set all coordinates for FlyingRobot

Samaev (changes requested):
good Job, nice robot factory

I guess it needs few corrections, look my comments please

Samaev (approved):
Perfect)

illiarerun (approved):
GJ

PR #866, opened 2022-06-23
cybersolution

Samaev:


yanaost (approved):
looks good)

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.maxLoadWeight >= this.currentLoad + cargo.weight) {
```
illiarerun, line 64: currentLoad is not a number, all you need is to check if `maxLoadWeight` is greater then the `cargo.weight`, and also just check if currentLoad doesn't have a `falsy` value
image

illiarerun (changes requested):
GJ, let's work on some fixes now:

src/robotFactory.js
Samaev: missunderstood at first
thanks, got it

illiarerun (approved):

PR #870, opened 2022-06-23
Solution

aedricson:


sancho325 (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords = {}, chipVersion, maxLoadWeight, currentLoad
+  ) {
+    super(name);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
```
jstmpelowycz, line 61: you'd rather add this
```
```
so that `undefined` won't ever be written

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
jstmpelowycz, line 65:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

jstmpelowycz (approved):

PR #881, opened 2022-06-23
solution was added

Bielichenko:


SpaceHook (approved):
GJ

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    Object.keys(coords).length > 1
+      ? this.coords = coords
+      : this.coords = {
+        x: 0, y: 0,
+      };
```
illiarerun, line 12: First of all, use default values for your function params
image

illiarerun: Second, don't repeat yourself, with assignment, transform it into something like this
```suggestion
    this.coords = Object.keys(coords).length > 1
      ? coords
      :  {
        x: 0,
        y: 0,
      };
```

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight = 0, currentLoad = 0) {
```
illiarerun, line 56: same logic as with html attributes
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight = 0,
    currentLoad = 0,
  ) {
```

illiarerun (changes requested):
Nice start, now let's work on some fixes:

Olexandr6 (approved):

PR #904, opened 2022-07-04
solution

silchyn:


src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
-class FlyingRobot {}
+    if (typeof coords === 'number') {
+      this.coords = {
+        x: 0,
+        y: 0,
+      };
 
-class DeliveryDrone {}
+      this.chipVersion = coords;
+    } else {
```
VETALEK, line 15: do you need this in some tests, because if you think that the user gonna set `chipVersion`, but not set `coords` you may just change the arguments order in constructor

VETALEK: because you can not use `coords` value as a `chipVersion`, make users set `coords` to `null`, if they want it to be the default value

VETALEK (changes requested):
GJ 🔥 
i am confused with your usage of `coords` and `chipVestion`

src/robotFactory.js
silchyn: Well, I thought I cannot just change the order because the tests' input will not change. Also, coords is the third argument, so I thought it will contain chipVersion if omitted even if I set the default value. Am I wrong?

silchyn: > because you can not use `coords` value as a `chipVersion`, make users set `coords` to `null`, if they want it to be the default value

Or maybe I can somehow specify the type which coords accepts?

silchyn: Understood)

PR #912, opened 2022-07-08
solution 1

TetianaKoval:


src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+    this.coords.x += step;
+  };
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
```
alexandra-protyanova, line 43: you've added `x & y` on the 40 line, no need to do it here

alexandra-protyanova (approved):
Well done!

src/robotFactory.js
TetianaKoval: ok, thank you)

PR #919, opened 2022-07-13
Solution

Sergii2705:


src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
 
-class FlyingRobot {}
+    this.coords.x = 'x' in this.coords
+      ? this.coords.x
+      : 0;
 
-class DeliveryDrone {}
+    this.coords.y = 'y' in this.coords
+      ? this.coords.y
+      : 0;
```
Ihor-Karpyn, line 15:
```suggestion
```

src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = { ...coords };
 
-class FlyingRobot {}
+    this.coords.x = 'x' in this.coords
+      ? this.coords.x
+      : 0;
```
Ihor-Karpyn, line 11:
```suggestion
```

src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = { ...coords };
```
Ihor-Karpyn, line 7:
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords.z = 'z' in this.coords
+      ? this.coords.z
+      : 0;
+  }
```
Ihor-Karpyn, line 65: the same
```suggestion
```

Ihor-Karpyn (changes requested):

src/robotFactory.js
Sergii2705: Then the question is why doesn't your system skip the "Nullish coalescing operator (??)" ?. It is more suitable here, it was for such cases that it was invented. I wanted to use it here, but I couldn't!??

VETALEK (approved):
🚀

PR #924, opened 2022-07-17
Solution

Lentyaeev:


src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    if (coords) {
+      if (!coords.x) {
+        coords.x = 0;
+      }
 
-class DeliveryDrone {}
+      if (!coords.y) {
+        coords.y = 0;
+      }
+      this.coords = coords;
```
VETALEK, line 17:
```suggestion
      this.coords = { ...coords };
```

let's not mutate the argument, in case it is used somewhere else

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    } else {
+      this.coords = {
+        x: 0, y: 0,
+      };
+    }
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
VETALEK, line 26:
```suggestion
    return `Robot: ${this.name}, Chip version: ${this.chipVersion},\
     Weight: ${this.weight}`;
```

VETALEK: this might help with linter 😉

VETALEK (changes requested):
Good start 😎

VETALEK (approved):
GJ 👍

PR #946, opened 2022-07-21
Solution

blcts:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
nikita-biriukov, line 44: you can not copy x and y coords
you can just add z coord

nikita-biriukov (approved):
good job 🔥 🔥

src/robotFactory.js
solaryasha: Agree with Nikita, `super` method already done it.
So, remove these lines

solaryasha (changes requested):

VETALEK (approved):
🚀

PR #960, opened 2022-07-24
solution

matvii1:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+  goRight(steps = 1) {
+    this.coords.x += steps;
+  }
+
+  goLeft(steps = 1) {
+    this.coords.x -= steps;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
Bohdan-Maksymiuk, line 31: you can use concatenation and pass parts on new lines to get rid of this error

Bohdan-Maksymiuk (approved):
good job

PR #972, opened 2022-08-08
solution

frontend112:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(...args) {
```
amitpatiljc, line 4: As per checklist you have to take parameters do not pass `...args`
constructor takes name, weight, coords, chipVersion and saves them

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+
+  getInfo() {
+    return (`Robot: ${this.name}, Chip version: ${
+      this.chipVersion}, Weight: ${this.weight}`);
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(...args) {
+    super(...args);
```
amitpatiljc, line 41: Here as well

amitpatiljc (changes requested):
Good job on passing all tests but go according to the checklist and make necessary changes

src/robotFactory.js
frontend112: Thank You for very fast response ;-)

amitpatiljc (approved):
Good Job!

PR #1000, opened 2022-08-23
solution

andredolgoleo:


src/robotFactory.js
```diff
@@ -1,10 +1,65 @@
...
+  }
+  goRight(a = 1) {
+    this.coords.x += a;
+  }
+  goLeft(a = 1) {
+    this.coords.x -= a;
+  }
+  getInfo() {
+    // eslint-disable-next-line max-len
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
alexandra-protyanova, line 27: u can combine strings using `+` operator

alexandra-protyanova (approved):
Good job!

src/robotFactory.js
andredolgoleo: Thank, i'll adopt it!

PR #1003, opened 2022-08-25
robot Factory solution

zinchenko-anastasiia:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
+  }
+
```
Andrii-Stiebalov, line 55: X and Y are inherited from BaseRobot. You only have to add the Z coordinate for FlyingRobot😉

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this;
+  }
+
```
Andrii-Stiebalov, line 20: image

It does **not** say that you need to return the object after movement.
If I missed something, then you don’t do a return for the Z coordinate.

image

Andrii-Stiebalov (changes requested):
Good job. But still needs some improvement 😉

Andrii-Stiebalov (approved):
GOOD JOB🚀👏

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    // eslint-disable-next-line max-len
```
danylo-onishchuk, line 32: delete it pls

danylo-onishchuk (changes requested):

danylo-onishchuk (approved):

PR #1009, opened 2022-08-25
Solution

Ohiiko:


Andrii-Stiebalov (approved):
!! Brillant job !! 👏🚀

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+  constructor(name, weight, coords = 0, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
```
AlinaMatuschak, line 12: by default your coords isn’t an object (it's zero). So you try get x and y from number 🥲

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+  };
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  };
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
AlinaMatuschak, line 54:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null,
  ) {
```

AlinaMatuschak (changes requested):
just fix the default value on BaseRobot. I expect that there will be an error if BaseRobot doesn’t have coords. If I wrong just write me here )

AlinaMatuschak (approved):

PR #1011, opened 2022-08-26
Solution

DimaKobzar7:


src/robotFactory.js
```diff
@@ -1,10 +1,102 @@
...
+  }
+
+  goRight() {
+    super.goRight();
+  }
+
+  goLeft() {
+    super.goLeft();
+  }
+
```
Andrii-Stiebalov, line 64: The `extends` keyword works using `prototypes`. It sets `FlyingRobot.prototype.[[Prototype]]` to `BaseRobot.prototype`. So **_if the method is not found_** in `FlyingRobot.prototype`, JavaScript takes it from `BaseRobot.prototype.` delete lines 49 to 60

src/robotFactory.js
```diff
@@ -1,10 +1,102 @@
...
+  hookLoad(cargo) {
+    if (this.currentLoad) {
+      return;
+    }
+
+    if (cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    }
+  }
+
```
Andrii-Stiebalov, line 95: You can do this with a single if. 😉
```
 hookLoad(cargo) {
      this.currentLoad = cargo;
    }
  }
```

Andrii-Stiebalov (changes requested):
Good job!! but needs a fix.

src/robotFactory.js
DimaKobzar7: Thank you for your comment. Before that, I thought that each method should be registered through super

Andrii-Stiebalov (approved):
Good job!!! 🚀👏

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+    if (step > 0) {
+      this.coords.y += step;
+    }
+  }
 
-class DeliveryDrone {}
+  goBack(step = 1) {
+    if (step > 0) {
+      this.coords.y -= step;
+    }
```
dellachaise, line 22: you do not need this check in all move methods in this task.
```suggestion
      this.coords.y -= step;
```

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  goUp(step = 1) {
+    if (step > 0) {
+      this.coords.z += step;
+    }
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
dellachaise, line 64: Make sure that your arguments are written in one line or each one starts from a new line.

Rewrite this way 👇 
```suggestion
  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```

Or this way 👇 
```suggestion
  constructor(
    name, 
    weight, 
    coords,
    chipVersion, 
    maxLoadWeight, 
    currentLoad
  ) {
```

dellachaise (changes requested):
Good job 👍 

Just add a small fixes according to comments 👇

dellachaise (approved):
Well done 👍

PR #1031, opened 2022-08-28
Implement Robot Factory Func

max-kravchenko:
Implement Robot Factory Func

### DIFF
--- a/src/robotFactory.js
+++ b/src/robotFactory.js
@@ -1,10 +1,80 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
+
+  goBack(step = 1) {
+    this.coords.y -= step;
+  }
+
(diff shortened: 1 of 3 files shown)


Anton-Liada (approved):
👌

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
dellachaise, line 45: you need to add there only `z` coord. Other you already have in `Base robot`, and you added it in `super` also.

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    this.currentLoad = currentLoad || null;
+    this.maxLoadWeight = maxLoadWeight;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && this.currentLoad === null) {
+      this.currentLoad = {
+        weight: cargo.weight,
+        description: cargo.description,
+      };
```
dellachaise, line 70:
```suggestion
      this.currentLoad = cargo;
```

dellachaise (changes requested):
Good job 👍 

Just add a little fixes according to comments 👇

src/robotFactory.js
max-kravchenko: Thanks!

dellachaise (approved):
🚀

PR #1040, opened 2022-08-29
solution

adrqq:


src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
+    this.coords = {
+      x: 0 + coords.x || 0,
+      y: 0 + coords.y || 0,
```
ibondrk, line 7: You don't need to write: `x: 0 + coords.x || 0`.You can do this in easyer way and just write it `x: coords.x || 0`. Doesn't have sense to make calculation when you check if coord exist

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
```
ibondrk, line 4: `coords = {}` doesn't have sense becouse you anyway creating object below. You can remove `= {}`
The same below in your constuctors

ibondrk (changes requested):
In general it's ok, but need some fixes)

dellachaise (approved):
Well done 👍

PR #1063, opened 2022-09-03
Initial solution

MihailKostov:


src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
+    const { x = 0, y = 0 } = position;
 
-class FlyingRobot {}
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class DeliveryDrone {}
+    this.coords = {
+      x: x,
```
qoa11a, line 12: You can use shorthand and write just prop name if its value name equals to prop name

qoa11a (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, position = {}, chipVersion) {
+    super(name, weight, position, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = position;
+
+    this.coords = {
+      x: x,
+      y: y,
+      z: z,
```
SanyaBratashchuk, line 47:
```suggestion
      x,
      y,
      z,
```

SanyaBratashchuk (approved):

PR #1064, opened 2022-09-04
Solution

zhenija:


src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
```
alexandra-protyanova, line 57: use super for all repetitive arguments

alexandra-protyanova (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
+    super(name, weight, coords, chipVersion);
```
Bohdan-Maksymiuk, line 63:
```suggestion
    currentLoad = null,
  ) {
    super(name, weight, coords, chipVersion);
```
this will have better readability

Bohdan-Maksymiuk (approved):

PR #1067, opened 2022-09-06
Develop

viktor-kotik:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.coords.x -= step;
+  };
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  };
+
+  getInfo() {
+    // eslint-disable-next-line
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
Bohdan-Maksymiuk, line 34: you may use concatenation instead of disabling rules

Bohdan-Maksymiuk (approved):

PR #1083, opened 2022-09-22
robot factory solution

TheVovchik:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    const { name, chipVersion, weight } = this;
+
+    return `Robot: ${name}, ` + `Chip version: ${chipVersion}, `
+      + `Weight: ${weight}`;
```
grafit-off, line 35: I think you don't need to use `+` in this case, at least the first plus is unnecessary when you used template quotes.

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    { x = 0, y = 0, z = 0 },
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(...arguments);
```
grafit-off, line 64: Wow, that's interesting, I hadn't thought of that

grafit-off (approved):
Good job

src/robotFactory.js
TheVovchik: thnx

Anna-Tkachenko (approved):
🔥

PR #1093, opened 2022-09-26
Solution

Shepeha-Andriy:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, { x = 0, y = 0, z = 0 },
+    chipVersion, maxLoadWeight, currentLoad) {
+    super(...arguments);
```
imaksweb, line 56: Creative approach, but I'm not sure that's valid though. As I can see `arguments` contains  `maxLoadWeight` and `currentLoad` as well, but we don't need to pass them to `super`. Looking forward to see what will mentor say about this case 🙂

imaksweb (approved):
Nice work 👍

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    this.coords = {
+      x: x,
+      y: y,
+    };
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return `Robot: ${this.name}, `
+      + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
naziamoff, line 17: try destructuring `this` here

src/robotFactory.js
naziamoff: I agree with Illia here, no need to pass the unused variables 👍

naziamoff (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1095, opened 2022-09-27
Solution

imaksweb:


Vladosik1337 (approved):
Good Job))

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      ...{
+        x: 0,
+        y: 0,
+      },
+      ...coords,
```
Denchepornyuk, line 13: you don't need the first object here, you can get `x` and `y` from coords object

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1097, opened 2022-09-28
feature added

ihorandrianov:


src/robotFactory.js
```diff
@@ -1,10 +1,102 @@
...
+      + `${this.chipVersion}, Weight: ${this.weight}`;
+  }
 
-class DeliveryDrone {}
+  goForward(y = 1) {
+    if (y > 0) {
+      this.coords.y += y;
+    }
+
+    return this;
```
DenysKleimenov, line 25: We don't need to return the object, just to change the coordinates inside of the object,

src/robotFactory.js
```diff
@@ -1,10 +1,102 @@
...
+    this.coords = {
+      x: 0,
+      y: 0,
+      z: 0,
+      ...coords,
+    };
+    this.chipVersion = chipVersion;
+  }
+
+  goUp(z = 1) {
```
DenysKleimenov, line 63: Name 'step' is clearer

src/robotFactory.js
```diff
@@ -1,10 +1,102 @@
...
+    super(name, weight, coords, chipVersion);
+    this.currentLoad = currentLoad;
+    this.maxLoadWeight = maxLoadWeight;
+  }
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    }
+
+    return this;
```
DenysKleimenov, line 92: we don't need to return the object

src/robotFactory.js
```diff
@@ -1,10 +1,102 @@
...
+  }
+  unhookLoad() {
+    this.currentLoad = null;
+
+    return this;
```
DenysKleimenov, line 97: same here

DenysKleimenov (approved):
Change variable for step in the method for z coords

silchyn (changes requested):
Task doesn’t meet all checklist requirements

Denchepornyuk (approved):

PR #1099, opened 2022-09-28
Solution

jojnajbude:


src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
 
-class FlyingRobot {}
+    if (Object.keys(coords).length < 2) {
+      const axises = ['x', 'y'];
 
-class DeliveryDrone {}
+      for (const axis of axises) {
+        if (!coords[axis]) {
+          coords[axis] = 0;
+        }
```
anastasiia-nahaiska, line 14: I think using the operator || is better for understanding, but it was useful for me to see another solution!))

anastasiia-nahaiska (approved):
GJ!))

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+      }
+    }
```
silchyn, line 16: Try to simplify this

silchyn (changes requested):

silchyn (approved):
Cool!

PR #1108, opened 2022-09-28
Solution

Andrew-Skoryk:
Implemented 3 robot classes

### DIFF
--- a/src/robotFactory.js
+++ b/src/robotFactory.js
@@ -1,10 +1,85 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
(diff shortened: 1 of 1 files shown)


src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  Moves(value = 1) {
+    return value < 0 ? 0 : value;
+  };
+
+  goRight(value) {
```
imaksweb, line 19: I would suggest you to change the name of variable `value` to `steps` for instance. Because, as for me, `value` is not descriptive name in this case 🙂

imaksweb (approved):
Good job 👍

src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  Moves(step = 1) {
```
naziamoff, line 15: Should it be capitalized?

naziamoff (approved):

PR #1112, opened 2022-09-29
Solution

Yeliseiev:


Maksym845 (approved):
NIce!

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return `Robot: ${this.name}, `
+    + `Chip version: ${this.chipVersion}, `
+    + `Weight: ${this.weight}`;
```
naziamoff, line 18: good, but you should destructure `this` here and it will look even better 👍

naziamoff (approved):

PR #1113, opened 2022-09-29
Task was solved

DenysKleimenov:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
```
OleksiiBlezniuk, line 4: Interesting solution =)

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+  constructor(name, weight, { x = 0, y = 0, z = 0 }, chipVersion) {
+    super(name, weight, {
+      x, y,
+    }, chipVersion);
+
+    this.coords = {
+      x,
+      y,
+      z,
+    };
```
OleksiiBlezniuk, line 47: You already have this object with x and y coordinates, just set z value here

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, {
+    x, y, z = 0,
+  }, chipVersion, maxLoadWeight, currentLoad = null) {
+    super(name, weight,
+      {
+        x, y, z,
+      }, chipVersion);
```
OleksiiBlezniuk, line 66: Would be more readable if you move arguments to new line

OleksiiBlezniuk, line 66:
```suggestion
  constructor(
    name,
    weight, 
    {
      x, y, z = 0,
    },
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
    super(
      name,
      weight,
      {
        x, y, z,
      },
      chipVersion
    );
```

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
OleksiiBlezniuk, line 73: Use shorthand

OleksiiBlezniuk, line 73:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

OleksiiBlezniuk (approved):

PR #1118, opened 2022-09-29
solution

artem-musii:


myroslav-diiak (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
+    this.coords = coords;
+    this.coords.x = this.coords.x || 0;
+    this.coords.y = this.coords.y || 0;
```
OleksiiBlezniuk, line 10: This is strange, first you assign object to `coords` property, then you assign this object properties again

Let's just assign new object to `coords` property and set values for `x` and `y` there

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
👍

PR #1119, opened 2022-09-29
solution was added

Maksym845:


Yeliseiev (approved):
Good job! ᕙ⁠(⁠⇀⁠‸⁠↼⁠‶⁠)⁠ᕗ

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  };
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+
+  getInfo() {
+    return `Robot: ${this.name}, `
++ `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
```
naziamoff, line 34: let's practice some destructuring here

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight,
+    coords, chipVersion, maxLoadWeight, currentLoad = null) {
```
naziamoff, line 55: think of a way to make it look more neat, that's not good if you break line like that

naziamoff (changes requested):
Well done! Just two small comments

Anna-Tkachenko (approved):
Good job!

PR #1120, opened 2022-09-29
solution to robot factory

dsydorenko534:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+    + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
+}
```
iPanda177, line 35:
```suggestion
  getInfo() {
    return `Robot: ${this.name}, `
    + `Chip version: ${this.chipVersion}, `
    + `Weight: ${this.weight}`;
  }
}
```

iPanda177 (approved):
GJ

src/robotFactory.js
naziamoff: it will look event better if we use destructuring here)

naziamoff (changes requested):
Well done, but use destructuring

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  };
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+    + `Chip version: ${this.chipVersion}, `
+    + `Weight: ${this.weight}`;
```
naziamoff, line 34: Looks better! But actually I meant that you should use **destructuring**. I approve but it is highly recommended to implement it by yourself
```suggestion
    const { name, chipVersion, weight } = this;

    return `Robot: ${name}, Chip version: ${chipVersion}, Weight: ${weight}`;
```

naziamoff (approved):

PR #1122, opened 2022-09-29
Solution v01

anton-iskryk:


artem-nasadchuk (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  getInfo() {
+    const { name, chipVersion, weight } = this;
```
OleksiiBlezniuk, line 32: Ok, it's not critical but it looks strange to destructure `this` as for me🙂

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
OleksiiBlezniuk, line 55: Too long line, better move this arguments to new lines

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    super(name, weight, coords, chipVersion);
+    this.currentLoad = currentLoad || null;
```
OleksiiBlezniuk, line 57:
```suggestion
    super(name, weight, coords, chipVersion);
    
    this.currentLoad = currentLoad || null;
```

OleksiiBlezniuk: That way is more readable

OleksiiBlezniuk (approved):
Good job! Left few comments about code style

PR #1123, opened 2022-09-29
done

iPanda177:


dsydorenko534 (approved):
GJ) you found such a good solution to this task😎

PR #1128, opened 2022-09-29
Solution

rstarovoit:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  };
+
+  getInfo() {
+    return 'Robot: ' + this.name + ', '
+      + 'Chip version: ' + this.chipVersion + ', '
+      + 'Weight: ' + this.weight;
+  }
```
grafit-off, line 35: Also, you can use destructuring and template strings here

grafit-off (approved):
GJ

Denchepornyuk (approved):

PR #1130, opened 2022-09-29
Solution

Rina07:


src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+        z,
+      },
+      chipVersion
+    );
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+  hookLoad(cargo) {
+    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = {};
```
korniienkokarina, line 78: you can omit this line, because you rewrite the property in the next line anyway

korniienkokarina (commented):
interesting solution with coords object, everything looks good, left one comment below

src/robotFactory.js
Rina07: fixed, much appreciated)

korniienkokarina (approved):

silchyn (approved):
Cool!

PR #1131, opened 2022-09-29
Solution

ArtemHulkoDev:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: 0,
+      y: 0,
+      ...coords,
+    };
```
glugovsky, line 12: Seems alright, but checklist point 2 recommends to use logic operator for setting initial values like this
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: 0,
+      y: 0,
+      z: 0,
+      ...coords,
+    };
```
glugovsky, line 47: If you follow the previous suggestion then you can set the initial value for only the z coordinate this way
```suggestion
    this.coords.z = coords.z || 0;
```

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null) {
```
glugovsky, line 61: You have many enough parameters to move every one of these to the separate line
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
```

glugovsky (approved):

src/robotFactory.js
OleksiiBlezniuk, line 12: Looks strange, why set values and then change it if `coords` passed, if we can just set coords once? 🙂

src/robotFactory.js
OleksiiBlezniuk, line 47: Same here

OleksiiBlezniuk: And also you already created this object in `super` constructor, just set coords.z value here

src/robotFactory.js
OleksiiBlezniuk, line 61: Too many arguments on one line, let's move them to new lines

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
OleksiiBlezniuk, line 68: You can use shorthand

OleksiiBlezniuk, line 68:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

OleksiiBlezniuk (changes requested):

src/robotFactory.js
ArtemHulkoDev: yeah, thank you for this info, fixed)

ArtemHulkoDev: I see, fixed it already, thanks)

src/robotFactory.js
ArtemHulkoDev: fixed)

src/robotFactory.js
ArtemHulkoDev: done)

src/robotFactory.js
ArtemHulkoDev: done)

silchyn (approved):
Cool!

PR #1136, opened 2022-10-01
created a robot factory

SerhiiRomaniukWD:


tatyana22 (approved):
GJ!

src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
```
Anna-Tkachenko, line 53: you should add here only z property
x and y will be set in BaseRobot constructor

src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+    };
+  }
+
+  goUp(step = 1) {
+    if (step > 0) {
+      this.coords.z += step;
+    }
```
Anna-Tkachenko, line 60: why do you need this conditions? you don't have them in the methods of BaseRobot class

Anna-Tkachenko (changes requested):

src/robotFactory.js
SerhiiRomaniukWD: yes, you are right

src/robotFactory.js
SerhiiRomaniukWD: we don't need it
i will fix it

Anna-Tkachenko (approved):
Good job!

PR #1139, opened 2022-10-01
Solution

andrew-satyr:


varkovroma (approved):
GJ

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
```
Anna-Tkachenko, line 44: you should add only z property here
x and y will be set in BaseRobot constructor

Anna-Tkachenko (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+      + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      ...this.coords,
```
Anna-Tkachenko, line 42: setting other properties is redundant here
you set x and y in BaseRobot constructor

Anna-Tkachenko (changes requested):

PR #1142, opened 2022-10-02
robotFactory

Arsen-Shpak:


MAKSOti (approved):
GJ 
but u need read checklist
[CODE KNOWLEDGE] - If in object variableName === keyName you can use short syntax:.

Anna-Tkachenko (approved):
Good job!

PR #1147, opened 2022-10-02
solution

poznianski:


Oleksandra-Shevchenko (approved):
LGTM 💯

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    // eslint-disable-next-line max-len
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
OleksiiBlezniuk, line 33: Well, instead of disabling linter, fix it's error 🙂  I believe you can split this string to new line

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.currentLoad = currentLoad || null;
+    this.maxLoadWeight = maxLoadWeight;
+  }
```
OleksiiBlezniuk, line 58:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad
  ) {
    super(name, weight, coords, chipVersion);
    
    this.currentLoad = currentLoad || null;
    this.maxLoadWeight = maxLoadWeight;
  }
```

OleksiiBlezniuk: When you have too many arguments in function, better to move every argument to new line, that way is more readable

Also would be nice to separate methods call (constructor in this case) and classes properties by empty line
It's much better for readability

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
OleksiiBlezniuk, line 61:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

OleksiiBlezniuk: Use shortcut

OleksiiBlezniuk (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion},`
+    + ` Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z || 0;
+  }
```
OleksiiBlezniuk, line 41:
```suggestion
  constructor(name, weight, coords, chipVersion) {
    super(name, weight, coords, chipVersion);
    
    this.coords.z = coords.z || 0;
  }
```

OleksiiBlezniuk (approved):

PR #1149, opened 2022-10-03
made classes for js_robot_factory

Hlazman:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight,
+    currentLoad = null) {
```
TarasHoliuk, line 57:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
```

correct formatting :arrow_up:

TarasHoliuk (approved):
GJ! Pay attention to code style (comment)

src/robotFactory.js
Hlazman: Understood, thanks for the advice

PR #1161, opened 2022-10-07
solution

GizemEce:


src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this;
```
hajnaloltyan, line 18: you can remove the 'return this',  it works without it too

hajnaloltyan (approved):
Great!

src/robotFactory.js
GizemEce: Done ✅ Thank you.

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
ihor-jpeg, line 63:
```suggestion
    if (
      this.currentLoad === null
      && cargo.weight <= this.maxLoadWeight
    ) {
```

ihor-jpeg (approved):

PR #1182, opened 2022-10-17
Solution

Ivanenko1402:


src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
Bohdan-Maksymiuk, line 48: you have already defined `x` and `y` in super constructor
so you can define only `z` here

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  };
+
+  hookLoad(cargo) {
+    if (this.maxLoadWeight >= cargo.weight) {
+      if (!this.currentLoad) {
```
Bohdan-Maksymiuk, line 70: you can use 1 if, instead of 2 and combine conditions

Bohdan-Maksymiuk (changes requested):
Good job, fix comments

polosanya (approved):
Good job!

PR #1204, opened 2022-10-29
solution

MaxFedak:


Illia-Skladnik (approved):
Looks good to me

src/robotFactory.js
```diff
@@ -1,10 +1,97 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0,
+  }, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
+    this.setCoords(coords);
```
Olexandr6, line 10: Do you really need this method? I would simplify it

src/robotFactory.js
```diff
@@ -1,10 +1,97 @@
...
+    this.coords.z += step;
+
+    return this;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
+  }, chipVersion, maxLoadWeight, currentLoad = null) {
```
Olexandr6, line 77: Format functions parameter in this case. Each parameter should be placed with new line

src/robotFactory.js
```diff
@@ -1,10 +1,97 @@
...
+    super(name, weight, coords, chipVersion, maxLoadWeight, currentLoad);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (Object.keys(cargo).length
```
Olexandr6, line 84: `Object.keys(cargo).length` - looks like this check is redundant.
Here you need only 2 checks

Olexandr6 (changes requested):
Almost done👍 Now, let's work on improvements:

Olexandr6 (approved):

PR #1210, opened 2022-11-04
Solution

Diana-Yrchyshyn:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+
+  goUp(step = 1) {
+    this.coords.z += step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight,
+    currentLoad = null) {
```
loralevitska, line 60:
```suggestion
  constructor(
    name, 
    weight,
    coords, 
    chipVersion, 
    maxLoadWeight,
    currentLoad = null
  ) {
```

loralevitska (changes requested):
almost done)

src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
+    this.coords.x -= step;
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  getInfo() {
+    return `Robot: ` + this.name + `, Chip version: ` + this.chipVersion
+    + `, Weight: ` + this.weight;
```
Anastasiia-Svintsova, line 34: It is better to use interpolation in such cases
```suggestion
    return `Robot: ${this.name}, ....`;
```

Anastasiia-Svintsova (approved):
Good! Check the comment, please

src/robotFactory.js
Diana-Yrchyshyn: How can I use interpolation if then the string size exceeds the limit?

qoa11a, line 34: It's not forbidden to include new lines inside a string. So you can just write it in 3 lines (one prop for line) instead of 1.

qoa11a (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+      + `Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
```
Anastasiia-Svintsova, line 35:
```suggestion
    return (
      `Robot: ${this.name}, ` +
      `Chip version: ${this.chipVersion}, ` +
      `Weight: ${this.weight}`
    );
```

Anastasiia-Svintsova (approved):

Anastasiia-Svintsova (approved):
I've already approved this work. The comment was for you to improve your code in the future. If the work has already been approved, you don't need to request a review again😉

PR #1219, opened 2022-11-07
solution

IOneStarI:


src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    const info = `Robot: ${this.name}, Chip version: ${this.chipVersion},`
```
masha-girya, line 21: you can return it immediately without constant

Olexandr6: +1

Olexandr6 (approved):
Good job 👍

PR #1220, opened 2022-11-07
robot factory MG

masha-girya:


src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
IOneStarI: you can set only z coords because x and y you already have)

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null
```
IOneStarI, line 55: my advice to don't write lots of parameters in single row)

IOneStarI (approved):
GJ!  but why your tests not going through ?

src/robotFactory.js
Olexandr6: +1 let's format parameters

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    this.currentLoad
+    = cargo.weight <= this.maxLoadWeight && this.currentLoad === null
+        ? { ...cargo }
+        : this.currentLoad;
```
Olexandr6, line 67: Would be better to use `if` statement here, cause your logic looks a bit complicated

src/robotFactory.js
Olexandr6: Each parameter should start with a new line

Olexandr6 (changes requested):
Almost done👍 Now, let's work on improvements:

Olexandr6 (approved):

PR #1230, opened 2022-11-08
solution

nazarko94:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
```
Andriy-Shymkiv, line 59:
```
  constructor(
      name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null
  ) {
```

Andriy-Shymkiv (approved):
gj

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
illiarerun, line 4: what if we try to pass undefined here as a coords? we would get an error, right. so please specify default param here, you can use plain object for that

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+    };
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return (
+      `Robot: ${this.name},`
+      + ` Chip version: ${this.chipVersion},`
+      + ` Weight: ${this.weight}`
+    );
```
illiarerun, line 20: use destructuring assgnment to avoid writing `this` everytime

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
illiarerun, line 38: same here with `coords`

illiarerun (approved):

PR #1232, opened 2022-11-08
develop

Ostapchuk-Y-O:


Kry-Gin (approved):
Well done! (👍≖‿‿≖)👍

src/robotFactory.js
```diff
@@ -1,10 +1,116 @@
...
+//   }
+//   coords: {
+//     x: number
+//     y: number
+//     z: number
+//   }
+// }
+
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
illiarerun, line 45: what if we try to pass `undefined` here as a `coords`? we would get an error, right. so please specify default param here, you can use plain object for that

src/robotFactory.js
```diff
@@ -1,10 +1,116 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null
+  ) {
```
illiarerun, line 97:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
```

illiarerun (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: `
+    + `${this.chipVersion}, Weight: ${this.weight}`;
```
mykola-koval, line 33: `const {name, chipVersion, weight} = this;`
and then
 
```
return `Robot: ${name}, Chip version: `
    + `${chipVersion}, Weight: ${weight}`;

mykola-koval (approved):

PR #1238, opened 2022-11-08
 js_robot_factory

MXMUK:


Oleksii-Siviolov (approved):
Looks good to me, nothing to add!

src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
```
illiarerun, line 7: what if we try to pass undefined here as a coords? we would get an error, right. so please specify default param here

src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}`
+    + `, Weight: ${this.weight}`;
```
illiarerun, line 38: use destructuring assgnment to avoid writing this everytime

src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
```
illiarerun, line 46: same here with `coords`

src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
...
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    }
+  }
+  unhookLoad() {
+    this.currentLoad = null;
+  }
```
illiarerun, line 83: add blank line between methods
```suggestion
  }
  
  unhookLoad() {
    this.currentLoad = null;
  }
```

illiarerun (changes requested):

illiarerun (approved):

PR #1240, opened 2022-11-08
solution

oksenyuk-valentyna:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+
+  getInfo() {
+    return 'Robot: ' + this.name
+    + ', Chip version: ' + this.chipVersion
+    + ', Weight: ' + this.weight;
+  };
```
sssolomia, line 35: maybe it will be better to use interpolation?

sssolomia (approved):
GJ!🚀

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
illiarerun, line 4: what if we try to pass `undefined` here as a `coords`? we would get an error, right. so please specify default param here, you can use plain object for that

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+    this.coords.x += step;
+  };
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+
+  getInfo() {
+    return `Robot: ${this.name},`
+    + ` Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
illiarerun, line 33: use destructuring assgnment to avoid writing `this` everytime

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+  goDown(step = 1) {
+    this.coords.z -= step;
+  };
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
```
illiarerun, line 56: use default params for that

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+  };
+
+  getInfo() {
+    return `Robot: ${this.name},`
+    + ` Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  };
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
illiarerun, line 38: again, what if `coord` undefined?

illiarerun (changes requested):

illiarerun (approved):

PR #1244, opened 2022-11-08
Solution

karyna-salnykova:


src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null
```
Yurii-Pozho, line 57: name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null   

For my opinion better to write in column

karyna-salnykova: I think there's no need, it's just params

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name},`
+      + ` Chip version: ${this.chipVersion},`
+      + ` Weight: ${this.weight}`;
+  }
```
illiarerun, line 36: [mateacademy.slack.com/archives/C041RC2F8JH/p1668011374756309](https://mateacademy.slack.com/archives/C041RC2F8JH/p1668011374756309)

src/robotFactory.js
illiarerun: i think there is, cause if your params list gets bigger than 80 lines, linter won't pass it and it'd be hard to read)

illiarerun (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
illiarerun, line 73:
```suggestion
    if (!this.currentLoad&& cargo.weight <= this.maxLoadWeight) {
```

illiarerun (approved):

PR #1246, opened 2022-11-08
Solution

referee1988:


Kry-Gin (approved):
Good job! (👍≖‿‿≖)👍

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
illiarerun, line 4: what if we try to pass `undefined` here as a `coords`? we would get an error, right. so please specify default param here, you can use plain object for that

illiarerun (changes requested):

PR #1255, opened 2022-11-09
add solution

AlexTsimba:


src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+
+    return this;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, load) {
```
Stepico, line 65: When we have a lot of arguments then it won't hurt if we write them, like, one per line
```
constructor(
    name,
    weight,
    coords,
    chipVersion
  ) {
  ```

Comment (AlexTsimba) [reply src/robotFactory.js]: great point, thanks!

File: src/robotFactory.js [hunk 2]
```diff
@@ -1,10 +1,98 @@
...
+    load
+  ) {
+    super(
+      name,
+      weight,
+      coords,
+      chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = load || null;
```
illiarerun, line 83: use default params for this

illiarerun (approved):

PR #1268, opened 2022-11-21
Implemented the solution for robot factory task

zynkd:


src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+  }
+
+  getInfo() {
+    return ('Robot: '
+            + this.name
+            + ', Chip version: '
+            + this.chipVersion
+            + ', Weight: '
+            + this.weight);
+  }
```
jstmpelowycz, line 46: looks prettier
```suggestion
  getInfo() {
	  const { name, weight, chipVersion } = this;
	
	  return `Robot: ${name}, `
	    + `chip version: ${chipVersion}, `
	    + `weight: ${weight}`;
  }
```

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+    currentLoad = null,
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
jstmpelowycz, line 85:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

jstmpelowycz (approved):
nicely done! left a few comments

PR #1270, opened 2022-11-23
test oK

diolna:


src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coord = {},
+    chipVersion, maxLoadWeight, currentLoad = null) {
```
loralevitska, line 72:
```suggestion
  constructor(
    name, 
    weight, 
    coord = {},
    chipVersion, 
    maxLoadWeight, 
    currentLoad = null
  ) {
```

loralevitska (changes requested):
great! need some improvements

src/robotFactory.js
diolna: fixed

diolna:
fixed

PR #1285, opened 2022-12-01
js_robot_factory

ysrckr:


src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
...
+    return this;
+  };
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
```
UmizDemud, line 51: x and y are being set by the parent constructor in super call, so probably you don't need those two

UmizDemud (approved):
Well done 👍

src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
...
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
meforoma, line 54: indeed method `super` sets x, y so you just need to add z:
`this.coords.z = coords.z || 0;`

src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
...
+  };
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
```
meforoma, line 76: if you multi-line, stick to the rule of aligned open and closing parenthesis
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null,
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
...
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && this.currentLoad === null) {
```
meforoma, line 84: use shorthand validation for falsy equivalents
unless you need to validate some particual falsy equivalent
```suggestion
    if (cargo.weight <= this.maxLoadWeight && !this.currentLoad) {
```

meforoma (changes requested):
few changes please))

meforoma (approved):
🎉

PR #1288, opened 2022-12-02
Solution

UmizDemud:


src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
+    coords = {
+      x: 0,
+      y: 0,
+    },
+    chipVersion
+  ) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
+    this.chipVersion = chipVersion;
```
meforoma, line 15: it feels easier to define default coords like this
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
    this.chipVersion = chipVersion;
```

then you won't need the next two `if (...) {...}`

src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
 
-class FlyingRobot {}
+    if (this.coords.x == null) {
+      this.coords.x = 0;
+    }
```
meforoma, line 19: you won't need these `ifs` if used logical OR operator when assigning (see above)

meforoma, line 19: you won't need these `ifs` if used logical OR operator when assigning (see above)

src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
+      y: 0,
+      z: 0,
+    },
+    chipVersion
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    if (this.coords.z == null) {
+      this.coords.z = 0;
+    }
```
meforoma, line 63: it's easier:
`this.coords.z = coords.z || 0;`

src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
+    maxLoadWeight,
+    currentLoad
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && this.currentLoad == null) {
```
meforoma, line 93: may be `=== null` is indeed required, but i suppose `!this.currentLoad` would work just fine.

meforoma: and yeah, better use strict comparison, not the loose one
🔴 `==`
🟢 `===`

meforoma (changes requested):
👏 
please read through the comments

meforoma (approved):
🎯

PR #1310, opened 2022-12-12
solution

kreisberg:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
valerii-derkach, line 55:
```suggestion
  constructor(
    name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null
   ) {
```

indentations deserve to be better here

valerii-derkach (changes requested):

src/robotFactory.js
kreisberg: ✔️

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null
+  ) {
```
valerii-derkach, line 61: every argument from new line
```suggestion
  constructor(
    name,
    weight, 
    coords, 
    chipVersion,
    maxLoadWeight, 
    currentLoad = null
  ) {
```
or all arguments in one line

```suggestion
  constructor(
    name, weight, coords, chipVersion, maxLoadWeight,  currentLoad = null
  ) {
```

valerii-derkach (changes requested):

src/robotFactory.js
kreisberg: got it)

valerii-derkach (approved):

PR #1311, opened 2022-12-12
Solution

s3rhiienko:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+      y: coords.y || 0,
+    };
 
-class DeliveryDrone {}
+    this.chipVersion = chipVersion;
+  };
+
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
```
loralevitska, line 18: add empty line between methods. Fix it everywhere)

```suggestion
  }
  
```

loralevitska (changes requested):
well done) need some improvements

vlad-demchuk (approved):

PR #1312, opened 2022-12-12
Solution

emxm:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+      + `Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
polosanya, line 39:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion,
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    this.coords.z += step;
+  };
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  };
+};
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
polosanya, line 55:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion, 
    maxLoadWeight, 
    currentLoad,
  ) {
```

polosanya (changes requested):
Almost done! Let's provide some fixes)

emxm:
After last commit, where I applied your recommendations, you requested changes second time but without any comments, but my code is passing autochecking. I don't know what I need to do(

polosanya:
> After last commit, where I applied your recommendations, you requested changes second time but without any comments, but my code is passing autochecking. I don't know what I need to do(

What do you mean? I requested only once, with comments)
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review 🙂

polosanya (approved):
Good job!

PR #1330, opened 2022-12-13
solution

Lun4trick:


ysrckr (approved):
LGTM 🤟🏻

src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
...
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return `Robot: ${this.name}, `
+    + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  };
```
naziamoff, line 18: better to keep it consistent
```suggestion
    return `Robot: ${this.name}, `
    + `Chip version: ${this.chipVersion},` 
    + `Weight: ${this.weight}`;
  };
```

src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.currentLoad = currentLoad || null;
+    this.maxLoadWeight = maxLoadWeight;
+  }
+
+  hookLoad(cargo) {
+    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 59: would be better to create a separate variable for that. If we'll need to somehow modify it - it will be much easier inside a variable than inside the if statement

naziamoff (approved):
LGTM, just a couple of suggestions

PR #1332, opened 2022-12-13
solution

mmahfuzi:


src/robotFactory.js
```diff
@@ -1,10 +1,101 @@
...
+
+    return this.coords.x;
+  }
+
+  getInfo() {
+    // These two variables are to avoid eslintmax-len error
+    const n = this.name;
+    const cV = this.chipVersion;
+
+    return `Robot: ${n}, Chip version: ${cV}, Weight: ${this.weight}`;
```
Lun4trick, line 44:
```suggestion
    return `Robot: ${n}, `
    + `Chip version: ${cV}, ` 
    + `Weight: ${this.weight}`;
```
This works too i think :)

Lun4trick (approved):
Very good :) 💯

src/robotFactory.js
mmahfuzi: Yes you are right :) I will change it. Thanks

src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this.coords.y;
+  }
+
+  goBack(step = 1) {
+    this.coords.y -= step;
+
+    return this.coords.y;
```
naziamoff, line 24: do you need this return?

src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+
+    return this.coords.x;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+      + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
naziamoff, line 41: let's move each concat part to the new line to keep it consistent

src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z || 0;
+  };
+  goForward(step = 1) {
+    super.goForward(step);
+  }
```
naziamoff, line 52: do you really need to repeat this logic?

src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 87: would look more simpler if the condition is separated into the variable

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #1341, opened 2022-12-14
Solution

CharAnagn:


grslckr (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
```
naziamoff, line 15: better to return this to make the chain call possible

naziamoff (approved):
🔥

PR #1342, opened 2022-12-14
Complete

SalehBUD:


grslckr (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
```
naziamoff, line 25: better to return `this` to make the chain call possible

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    const n = this.name;
+    const c = this.chipVersion;
+    const w = this.weight;
```
naziamoff, line 42: you can just use destructuring to get variables `name`, `chipVersion`, `weight`

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 83: let's use shortcut for checking whether the boolean is falsy
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

naziamoff (approved):
looks good, but there are some suggestions

PR #1347, opened 2022-12-14
solution

grslckr:


nurdanokcu (approved):
:+1:

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return 'Robot: ' + this.name
+    + ', Chip version: ' + this.chipVersion
+    + ', Weight: ' + this.weight;
```
naziamoff, line 34: looks simpler this way
```suggestion
    return `Robot: ${this.name}`
    + `Chip version: ${this.chipVersion}`
    + `Weight: ${this.weight}`;
```

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
```
naziamoff, line 17: better to return `this` to make the chain call possible

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 62: let's use shortcut for checking whether the boolean is falsy
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

naziamoff (approved):
Looks good, just a couple of suggestions

PR #1350, opened 2022-12-14
Solution

TetKopot:


src/robotFactory.js
```diff
@@ -1,10 +1,69 @@
...
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x,
+      y,
+    };
```
RedkoSerhii, line 11: CL 2: use logic operator || if you want safely add some initial values.
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };

src/robotFactory.js
```diff
@@ -1,10 +1,69 @@
...
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords,
+    chipVersion, maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
```
RedkoSerhii, line 54: CL 2

RedkoSerhii (changes requested):

src/robotFactory.js
TetKopot: IMHO in this way is harder to understand why we have 0

valerii-derkach: both options ok

src/robotFactory.js
```diff
@@ -1,10 +1,69 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords,
+    chipVersion, maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
```
valerii-derkach, line 52:
```suggestion
  constructor(
   name, 
   weight, 
   coords,
   chipVersion, 
   maxLoadWeight, 
   currentLoad = null
  ) {
    super(name, weight, coords, chipVersion);
```

indentations deserve to be better here

valerii-derkach (changes requested):

PR #1356, opened 2022-12-15
solution

victoria-kovalenko:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+
+    this.currentLoad = currentLoad || null;
+  }
```
viktoriia-kravchenko, line 63: I suggest moving setting null as a default value to currentLoad higher and write something like that: constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null).

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = 0, chipVersion) {
```
viktoriia-kravchenko, line 4: We can substitute coords = 0 here with  { x = 0, y = 0 } to increase readability.

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
```
viktoriia-kravchenko, line 11: The simplified version would be this.coords = { x, y, }, since we have already set the default value to coordinates.

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x,
+      y: coords.y,
+      z: coords.z || 0,
+    };
+  }
```
viktoriia-kravchenko, line 46: It's possible to omit to mention x and y here and define only z.

viktoriia-kravchenko (changes requested):
GJ! Only minor improvements can be made.

PR #1361, opened 2022-12-15
Robot Factory

Mohameddahi:


src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+    }
+
+    return this;
+  }
+
+  goDown(step = 1) {
+    if (step > 0) {
+      this.coords.z -= step;
+    }
+  }
```
ysrckr, line 71:
```suggestion
    }
    return this;
  }
```
`return this` was not required in the task but you have in other movements so it needs to be consistent.

ysrckr (changes requested):
Just a small change.

Mohameddahi:
> Just a small change.

@ysrckr thank you for your review.

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipversion, maxLoadWeight,
+    currentLoad = null) {
+    super(name, weight, coords, chipversion);
```
naziamoff, line 67: You should keep all the args in one line or each in a separate one
```suggestion
class DeliveryDrone extends FlyingRobot {
  constructor(
    name, 
    weight, 
    coords, 
    chipversion, 
    maxLoadWeight,
    currentLoad = null,
  ) {
    super(name, weight, coords, chipversion);
```

src/robotFactory.js
naziamoff: agree with @ysrckr

naziamoff (approved):
🔥

PR #1362, opened 2022-12-15
solution

kenvastolord:


src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = 0, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
ysrckr, line 41: You are already calling these with super so they are unnecessary.

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = 0, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
ysrckr, line 10: You are setting it 0 if it is `undefined` in `constructor`.  So you can delete 0 from `constructor` arguments.

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight > this.maxLoadWeight) {
+      return;
+    }
+    this.currentLoad = this.currentLoad || cargo;
```
ysrckr, line 71:
```suggestion
   if (cargo.weight <= this.maxLoadWeight && !this.currentLoad) {
      this.currentLoad = cargo;
    }
```
Shorter and takes in the consideration if there is a current load.

ysrckr (changes requested):

src/robotFactory.js
kenvastolord: if I don't call the coords X and Y in those lines, the test show me an error coords are not defined

ysrckr: I think you found the way to do it without getting an error.

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    }
+  }
+
+  unhookLoad(cargo) {
+    if (cargo <= this.maxLoadWeight) {
+      this.currentLoad = null;
+    }
```
ysrckr, line 71: You don't need this if statement.

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
ysrckr, line 63:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```
This is shorter

ysrckr (approved):
I left a few suggestions for you to consider.

src/robotFactory.js
kenvastolord: got it bro, Thanks a lot!

src/robotFactory.js
kenvastolord: yep, it works without the if, thanks!

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
```
naziamoff, line 58:
```suggestion
    maxLoadWeight,
    currentLoad = null
  ) {
    super(name, weight, coords, chipVersion);
    this.maxLoadWeight = maxLoadWeight;
```

naziamoff (approved):
Well done 🔥

PR #1363, opened 2022-12-15
solution

lukacsja:


ysrckr (approved):
Looks good 👍🏻

naziamoff (approved):
LGTM!
But in movement methods would be better if you made the chain call possible

PR #1364, opened 2022-12-15
Develop

TheOneThatKnowsNothing:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
OlggaJul, line 42: i think you need only z-coord here

OlggaJul (approved):

src/robotFactory.js
ivan-kadykalo: You are right. Let's fix it

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #1366, opened 2022-12-15
Solution

Iksich:


ysrckr (approved):
LGTM 👍🏻

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+      y: coords.y || 0,
+    };
 
-class DeliveryDrone {}
+    this.chipVersion = chipVersion;
+  }
+
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
```
naziamoff, line 18: better to return this to make the chain call possible

naziamoff (approved):

PR #1372, opened 2022-12-16
Solution

sercansimsek:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
ysrckr, line 63: You can change `this.currentLoad === null` to `!this.currentLoad`

ysrckr (approved):
Looks good just minor suggestion that might help you in future too.

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
 
-class DeliveryDrone {}
+    this.chipVersion = chipVersion;
+  };
+  goForward(step = 1) {
+    this.coords.y += step;
+  };
```
naziamoff, line 17: better to return this to make the chain call possible

src/robotFactory.js
naziamoff: agree with @ysrckr

naziamoff (approved):
GJ, just a couple of comments

PR #1376, opened 2022-12-16
Develop

dolnaro:


Lun4trick (approved):
looks good ! 👍 maybe its better to put those initial values in to this.coords in BaseRobot just as you did with the 'z' coordinate.

meforoma (approved):
well done 👍

PR #1379, opened 2022-12-17
Finished

JzSol:


src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
```
ysrckr, line 17:
```suggestion
  goForward(step = 1) {
    this.coords.y += step;
    
    return this;
  }
```
If you add `return this` to movement functions, you can chain them.

ysrckr (changes requested):
Looks good. Just one easy suggestion. If you do that it would be perfect.

meforoma (approved):
sweet 👍

PR #1381, opened 2022-12-18
solution

Alireza-Mohammadhossein:


src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return `Robot: ${this.name}, `
+    + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
meforoma, line 17: no need for `this.` - all three are input arguments
```suggestion
    return `Robot: ${name}, `
    + `Chip version: ${chipVersion}, Weight: ${weight}`;
```
🤔 prob would even fit the 80 chars

meforoma (approved):
valid ✅ 
just one tiny comment to consider

PR #1389, opened 2022-12-23
Solution robot factory

katerinalex:


src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    this.coords.x += step;
+  };
+
+  getInfo() {
+    let result = `Robot: ${this.name}, `;
+
+    result += `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+
+    return result;
+  }
```
Anastasiia-Svintsova, line 37: Looks a little bit strange 

```suggestion
    return `Robot: ${this.name}, `
      + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
  }
```

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+
+    if (!currentLoad) {
+      this.currentLoad = null;
+    } else {
+      this.currentLoad = currentLoad;
+    }
```
Anastasiia-Svintsova, line 65: You could use a ternary operator here or just set currentLoad as null by default in the constructor
```suggestion
   this.currentLoad = currentLoad ? currentLoad : null;
```

Anastasiia-Svintsova (changes requested):

src/robotFactory.js
katerinalex: Лінтер не видає помилку: 
this.currentLoad = currentLoad ? currentLoad : null;
Unnecessary use of conditional expression for default assignment.

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    this.coords.z -= step;
+  };
+
+  goUp(step = 1) {
+    this.coords.z += step;
+  };
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
polosanya, line 54:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion, 
    maxLoadWeight, 
    currentLoad,
  ) {
```

polosanya (approved):
Good job!

PR #1402, opened 2023-01-04
Solution

anna-akhalmosulishvili:


src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+      if (Object.keys(this.coords).length === 1) {
+        if (Object.keys(this.coords)[0] === 'x') {
+          this.coords = coords;
+          this.coords.y = 0;
+        } else {
+          this.coords.x = 0;
+          this.coords.y = Object.values(this.coords)[0];
+        }
+      }
+    }
```
DarkMistyRoom, line 26: `this.coords = {
  x: coords.x || 0,
  y: coords.y || 0,
};`

DarkMistyRoom: you can replace your huge solution with this and get the same result

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion,);
+
+    this.currentLoad = (!currentLoad)
+      ? null
+      : currentLoad;
```
DarkMistyRoom, line 78:
```suggestion
  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null) {
    super(name, weight, coords, chipVersion,);

    this.currentLoad = currentLoad;
```

src/robotFactory.js
DarkMistyRoom: fix the constructor in the FlyingRobot too

DarkMistyRoom (changes requested):

DarkMistyRoom (approved):

PR #1416, opened 2023-01-10
solution

igor-kashchenko:


OlhaYevstifieieva (approved):
Nice job 👋🏻

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this;
+  }
+
+  goBack(step = 1) {
+    this.coords.y -= step;
+
+    return this;
```
alexandra-protyanova, line 24: In this task you shouldn't return `this`, methods won't be called in chain

alexandra-protyanova (approved):

PR #1426, opened 2023-01-12
Solution

chermak-oleh:


PolinaSukhorukova (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
Anna-Tkachenko, line 55: will be better to move each argument to the new line to increase code readability

Anna-Tkachenko (approved):

PR #1440, opened 2023-01-16
Solution

maasiimka:


src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, null, chipVersion);
+
+    this.coords = {
+      x: 0,
+      y: 0,
+      z: 0,
+      ...coords,
```
GarryRocksman, line 48: here you don't need to pass null to BaseRobot constructor but just need to add z to coords.

GarryRocksman (approved):
GJ🔥 But needs a little fixes

src/robotFactory.js
ArturErmolenko: agree

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
```
ArturErmolenko, line 68:
```suggestion
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
    ) {
```

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #1443, opened 2023-01-16
solution

Oleksandr-Maly:


src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+  constructor(name, weight, coords = {}, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = coords;
+
+    this.coords = {
+      x,
+      y,
+      z,
+    };
```
GarryRocksman, line 49: you don't need this variable here because coords is already an object, here you just need to add z coord

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
+    const { x = 0, y = 0 } = coords;
```
GarryRocksman, line 5: you don't need this variable here because coords is already an object

GarryRocksman (approved):
GJ🔥 But needs a little fixes

src/robotFactory.js
ArturErmolenko, line 49: follow DRY

ArturErmolenko (changes requested):
GJ 👍  Need to fix comments

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    this.coords = {
+      x,
+      y,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return `Robot: ` + this.name + `, Chip version: `
```
Anna-Tkachenko, line 16: will be better to use interpolation instead of concatenation here

Anna-Tkachenko (approved):

PR #1457, opened 2023-01-16
solved task

Serghii29:


src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return ('Robot: ' + this.name + ',' + ' '
```
victorionov, line 32: Maybe its better to make it with "Template string". 
Something like:
 `Robot: ${this.name}.............`

victorionov (approved):

src/robotFactory.js
ArturErmolenko: agree use backticks

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight
+        && this.currentLoad === null) {
```
ArturErmolenko, line 64:
```suggestion
    if (cargo.weight <= this.maxLoadWeight
        && this.currentLoad === null
        ) {
```

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #1458, opened 2023-01-16
 solution

OlehSharha:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+  getInfo() {
+    // eslint-disable-next-line max-len
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
Avichez, line 33: You can just use here concatenation like this.
```suggestion
    return `Robot: ${this.name},`
    + ` Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```

Avichez (approved):
Great solution 👍 
Just added a small comment.

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+  getInfo() {
+    // eslint-disable-next-line max-len
```
ArturErmolenko, line 32: don't disable eslint
```suggestion
```

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name,
+    weight, coords, chipVersion, maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
```
ArturErmolenko, line 56:
```suggestion
  constructor(
    name,
    weight,
    coords, 
    chipVersion,
    maxLoadWeight,
    currentLoad = null,
  ) {
    super(name, weight, coords, chipVersion);
```

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #1463, opened 2023-01-16
Added solution

burovetskyi:


GarryRocksman (approved):
GJ🔥

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
```
Anna-Tkachenko, line 58: you can also set null as default value while describing function arguments

Anna-Tkachenko (approved):

PR #1474, opened 2023-01-26
robotFactory was made

SamirMamedov1010:


src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
```
tetiana-ivanchenko, line 4:
```suggestion
  constructor(name, weight, coords = { x = 0, y = 0 }, chipVersion) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x,
+      y,
+    };
```
tetiana-ivanchenko, line 11:
```suggestion
    this.coords = coords;
```

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+
+  getInfo() {
+    return `Robot: ${this.name},`
+      + ` Chip version: ${this.chipVersion},`
+      + ` Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, { x, y, z = 0 }, chipVersion) {
```
tetiana-ivanchenko, line 39: the same here

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+  goDown(z = 1) {
+    this.coords.z -= z;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    { x, y, z },
```
tetiana-ivanchenko, line 64: the same

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (changes requested):
GJ 🔥

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+  constructor(name, weight, { x, y, z = 0 }, chipVersion) {
+    super(
+      name,
+      weight, {
+        x,
+        y,
+        z,
+      }, chipVersion);
+
+    this.coords.z = z;
```
TarasHoliuk, line 48: since you do it anyway, it may look this way:
```suggestion
    this.coords.z = z || 0;
```
and no need to do destructuring assignment in the `constructor` arguments, and pass the new object in `super`

do the same in the classes above

TarasHoliuk: It'll be more readable

TarasHoliuk (changes requested):
Please take into account the comments
If you have questions feel free to ask in chat

src/robotFactory.js
SamirMamedov1010: Hi, thank you for check, but i already asked about this in chat and we decided to keep my option with Tetiana https://mate-parttime.slack.com/archives/C03UQS59NRY/p1674823760016929

SanyaBratashchuk (approved):

PR #1475, opened 2023-01-27
Solution

ruslan-yarosh:


src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x ? coords.x : 0,
```
ViktorHavryliuk, line 9: You can use ?? or || operators here as well
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
```suggestion
      x: coords.x ?? 0,
```

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    this.currentLoad = this.currentLoad === null
+      && cargo.weight <= this.maxLoadWeight
+      ? cargo
+      : this.currentLoad;
```
ViktorHavryliuk, line 68: This way it is easier to read and understand
```suggestion
    if(!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
      this.currentLoad = cargo
    }
```

ViktorHavryliuk (changes requested):
Good work, only few small suggestions

src/robotFactory.js
ruslan-yarosh: Thank you for an advise. I used || here because the linter test fails with ?? operator. But ?? operator is very usefull I will keep it on mind

SanyaBratashchuk (approved):

PR #1476, opened 2023-01-27
Solution

IrynaK-a:


src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+
+  getInfo() {
+    return (
+      'Robot: '
+      + this.name
+      + ', Chip version: '
+      + this.chipVersion
+      + ', Weight: '
+      + this.weight
+    );
```
agniya-i, line 41: you can also use template strings, for better readability 

```suggestion
     return `Robot: ${this.name}, `
      + `Chip version: ${this.chipVersion}, `
      + `Weight: ${this.weight}`;
```

agniya-i (approved):
Good job!

PR #1479, opened 2023-01-30
Robot factory was built

Rio1391:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    super(name, weight, chipVersion, coords);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight
+        && this.currentLoad === null
+    ) {
```
Volodymyr-Melnyk-Fron10der, line 65: I understand there might be eslint errors, but anyway, try to make it less weird, example:
```suggestion
    const isWeightAllowed = cargo.weight <= this.maxLoadWeight;
    
    if (isWeightAllowed && !this.currentLoad) {
```

Volodymyr-Melnyk-Fron10der (approved):
Great job! Just make you conditions more readable

PR #1485, opened 2023-02-05
done

jaxxxon78:


src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  getInfo() {
+    /* eslint-disable */
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
```
IvanFesenko, line 37: You can use destructuring here and remove` eslint-disable`
```suggestion
  getInfo() {
    const {name, chipVersion, weight} = this;
    
    return `Robot: ${name}, Chip version: ${chipVersion}, Weight: ${weight}`;
  }
```

src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
 
-class DeliveryDrone {}
+  goBack(step = 1) {
+    this.coords.y -= step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
```
IvanFesenko, line 29: Add new line between methods and after constructor

IvanFesenko (changes requested):
Good job. Please fix 2 small code style issues

qoa11a (approved):

PR #1489, opened 2023-02-06
robot factory created

kirill-vypirovskyi:


lilia-obushenko (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
OleksiiBlezniuk, line 10: FYI if coords object will not be passed to this constructor - you'll get an error Cannot read property 'z' from undefined. In newer JS specification this issue can be fixed using [Optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return 'Robot: ' + this.name
+      + ', Chip version: ' + this.chipVersion
+      + ', Weight: ' + this.weight;
+  // I can't do it with `` because of max line length
```
OleksiiBlezniuk, line 19: Actually you can split this string and then add two strings

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return 'Robot: ' + this.name
+      + ', Chip version: ' + this.chipVersion
+      + ', Weight: ' + this.weight;
```
OleksiiBlezniuk, line 18:
```suggestion
    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, `
      + `Weight: ${this.weight}`
```

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
OleksiiBlezniuk, line 55: Looks like this line is too long, let's move params to new lines

OleksiiBlezniuk (changes requested):
GJ! There are few small issues to fix

Zelinskyi-Serhii (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
```
OleksiiBlezniuk, line 63: or assign null to `currentLoad` as a default value

OleksiiBlezniuk (approved):

PR #1491, opened 2023-02-07
Solution

Vasyl-Pavlenko:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    const robot = `Robot: ${this.name}`;
+    const Chip = `Chip version: ${this.chipVersion}`;
+    const Weight = `Weight: ${this.weight}`;
+
+    return `${robot}, ${Chip}, ${Weight}`;
```
Viktor-Kost, line 36: you can use as one template string

Viktor-Kost (approved):
Nice!)

PR #1492, opened 2023-02-07
task was solved

vanyachyzh:
I use concatanation in getInfo method because of linter. It doesn't allow to have strings with over 80 characters in one row so I can't apply interpolation.

### DIFF
--- a/src/robotFactory.js
+++ b/src/robotFactory.js
@@ -1,10 +1,70 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
+    this.chipVersion = chipVersion;
+  }
 
-class FlyingRobot {}
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
 
-class DeliveryDrone {}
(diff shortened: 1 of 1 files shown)


src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
```
tetiana-ivanchenko, line 9:
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return 'Robot: ' + this.name + ', Chip version: '
+    + this.chipVersion + ', Weight: ' + this.weight;
```
tetiana-ivanchenko, line 31:
```suggestion
    return 'Robot: ' + this.name + ', Chip version: '
      + this.chipVersion + ', Weight: ' + this.weight;
```

tetiana-ivanchenko (changes requested):

PR #1497, opened 2023-02-10
done

l4st1m0za:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
```
ViktorHavryliuk, line 9: This will cause error if coords are not passed, please add optional chaining operator
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
image

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z || 0;
```
ViktorHavryliuk, line 40: same here

ViktorHavryliuk (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,95 @@
...
+      this.coords = {
+        x: 0,
+        y: 0,
+      };
+    } else {
+      this.coords = {
+        x: (coords.x === undefined) ? 0 : coords.x,
+        y: (coords.y === undefined) ? 0 : coords.y,
+      };
+    };
```
Volodymyr-Melnyk-Fron10der, line 18: Did you check the link previous mentor left for you?)
All you had to do is: 
```
 this.coords = {
    x: coords?.x || 0,
    y: coords?.y || 0,
 }
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, I think you didn't clearly understand what to do)

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: x,
+      y: y,
+    };
```
Volodymyr-Melnyk-Fron10der, line 11: If you have same key/value you can write it once
```suggestion
    this.coords = { x, y };
```

PR #1510, opened 2023-02-17
Solution

taras-nakonechnyi:


src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
```
OleziO, line 21: Add condition. If the step is greater than zero - do step, if no - do nothing. Check other similar cases.

OleziO: Return this for chain execution of methods. Check other similar cases.

src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion
```
OleziO, line 8: Add coords to constructor params.

OleziO (changes requested):

src/robotFactory.js
taras-nakonechnyi: What do you mean? In line 7, I add it.

tetiana-ivanchenko (approved):

PR #1513, opened 2023-02-19
Solution

lilia-obushenko:


src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
kuznetsova-anastasiia, line 4: you should put each prop on the next line if there are more than 2. Also you can use destructuring in coords
```suggestion
  constructor(
    name,
    weight,
    { x = 0, y = 0 },
    chipVersion
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
kuznetsova-anastasiia, line 10: also here it would be x and y with destructuring

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+
+    return this;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
kuznetsova-anastasiia, line 70:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad
  ) {
```

kuznetsova-anastasiia (approved):

src/robotFactory.js
OleksiiBlezniuk: Agree, would be nice to use destructuring

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+  }
+
+  getInfo() {
+    return `Robot: ${
+      this.name
+    }, Chip version: ${
+      this.chipVersion
+    }, Weight: ${
+      this.weight
+    }`;
```
OleksiiBlezniuk, line 46: Overkill, it became harder to read

OleksiiBlezniuk, line 46:
```suggestion
    return `Robot: ${this.name}, Chip version: ${this.chipVersion}`
      + `Weight: ${this.weight}`;
```

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z || 0;
```
OleksiiBlezniuk, line 53: FYI if coords object will not be passed to this constructor - you'll get an error Cannot read property 'z' from undefined. In newer JS specification this issue can be fixed using [Optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

src/robotFactory.js
OleksiiBlezniuk: Agree, fix it please, this line is too long

OleksiiBlezniuk (changes requested):

PR #1514, opened 2023-02-20
robotFactory solution

kateryna-babina:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
kirill-vypirovskyi, line 56: Better do each parameter from new line

kirill-vypirovskyi (approved):
LGFM

package.json
```diff
@@ -16,14 +16,13 @@
   "license": "GPL-3.0",
```
OleksiiBlezniuk, line 16: Please don't add such files to commits

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
```
OleksiiBlezniuk, line 11: FYI if coords object will not be passed to this constructor - you'll get an error Cannot read property 'z' from undefined. In newer JS specification this issue can be fixed using [Optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.coords.x -= step;
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+      + `Chip version: ${this.chipVersion}, `
```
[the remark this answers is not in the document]
OleksiiBlezniuk: Agree, fix it

src/robotFactory.js
OleksiiBlezniuk: Agree, if you moving params to new lines - move all of them

OleksiiBlezniuk (changes requested):

src/robotFactory.js
OleksiiBlezniuk: And you can use destructuring assignment for theses params

sergii-nosachenko (approved):
Good work 👍

PR #1515, opened 2023-02-20
Solution

GUSILLUS:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords.z = coords.z || 0;
+  }
+
+  goUp(step = 1) {
+    this.coords.z += step;
+  };
+  goDown(step = 1) {
```
OlEzhA0, line 48:
```suggestion
  };

  goDown(step = 1) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  };
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  };
+
+  getInfo() {
+    return `Robot: ${this.name},`
+     + ` Chip version: ${this.chipVersion},`
+     + ` Weight: ${this.weight}`;
```
OlEzhA0, line 34: why do you use string concatenation and backquotes?

PR #1518, opened 2023-02-21
Solution

vlad-ptashnichenko:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
```
OleziO, line 16: I recommend adding a check that the step is greater than zero and returning `this`, for the chain calling of these functions. Check other similar cases.

OleziO (approved):

PR #1519, opened 2023-02-22
solution

RuslanLiapin:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+    this.weight = weight;
+
+    this.chipVersion = chipVersion;
+  }
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${
+      this.chipVersion}, Weight: ${this.weight}`;
+  }
+  goForward(step = 1) {
+    this.coords.y += step;
```
OleziO, line 21: I recommend you add `if(step > 0)` and return `this` (for chain calling functions). Check other similar cases.

OleziO (approved):

PR #1520, opened 2023-02-22
Solution

stanislavcodes:


src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+      x,
+      y,
+    };
+  }
 
-class DeliveryDrone {}
+  getInfo() {
+    return (
+      `Robot: ${this.name}, `
+      + `Chip version: ${this.chipVersion}, `
```
kirill-vypirovskyi, line 18: You can keep 17 and 18 on the same line, there is anough place

kirill-vypirovskyi (approved):
LGFM

OleksiiBlezniuk (changes requested):

illiarerun (approved):

PR #1524, opened 2023-02-24
Solurion

timoxa2t:


src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords = {},
```
OleksiiBlezniuk, line 7: I guess if we using default value for param - better to assign valid value

{ x: 0, y: 0 }

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    chipVersion
+  ) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
+    this.coords.x = this.coords.x || 0;
+    this.coords.y = this.coords.y || 0;
```
OleksiiBlezniuk, line 14: Why to set value for each property of object if you can set an object?

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    const { name, chipVersion, weight } = this;
```
OleksiiBlezniuk, line 35: IMO destructuring of `this` object looks a bit weird

OleksiiBlezniuk: But you may leave it as it is now

OleksiiBlezniuk (changes requested):

src/robotFactory.js
timoxa2t: if coords isn't undefined I don't know what keys it has. So I'm making sure that I have x & y in my coords

src/robotFactory.js
```diff
@@ -1,10 +1,106 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords = {
+      x: 0,
+      y: 0,
+    },
```
sergii-nosachenko, line 10:
```suggestion
    coords,
```
Unnecessary, you do this when destructurize coords.

src/robotFactory.js
```diff
@@ -1,10 +1,106 @@
...
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = coords;
+
+    this.coords = {
+      x,
+      y,
+      z,
+    };
```
sergii-nosachenko, line 66: You need to add only z property.

sergii-nosachenko (changes requested):

src/robotFactory.js
timoxa2t: and if coords isn't an object?

sergii-nosachenko: Of course, it can be anything. But what if I say that `coords.x` can possibly be not a number? Or cargo can possibly be a string and doesn't have the weight prop.
That is why we are using Typescript 😉
In this task, we can be sure that coords are always objects to simplify checks.

src/robotFactory.js
```diff
@@ -1,10 +1,95 @@
...
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
```
sergii-nosachenko, line 81: Cargo can possibly be not an object. But this task guarantees that you'll get a proper object with a weight prop.

src/robotFactory.js
```diff
@@ -1,10 +1,95 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords = {},
```
sergii-nosachenko, line 7:
```suggestion
    coords,
```

Coords are always an object. No need to assign it to an empty object.

src/robotFactory.js
```diff
@@ -1,10 +1,95 @@
...
+
+    return `Robot: ${name}, Chip version: ${chipVersion}, Weight: ${weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords = {},
```
sergii-nosachenko, line 48: Same here

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1529, opened 2023-02-26
sol

yurii-filiushkin:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords,
+    chipVersion, maxLoadWeight, currentLoad = null) {
```
TarasHoliuk, line 55:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
```
TarasHoliuk, line 15: it will move backward with `goForvard(-1)` - handle this case

TarasHoliuk (changes requested):

src/robotFactory.js
yurii-filiushkin: Like this?

goForward(step = 1) {
    if (step < 0) {
      this.coords.y -= step;
    }
    this.coords.y += step;
  }

src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
+    if (step < 0) {
+      this.coords.y -= step;
+    }
+    this.coords.y += step;
```
tetiana-ivanchenko, line 19: you just need to check if your step is above 0 because -1 is not step forward

tetiana-ivanchenko (changes requested):

src/robotFactory.js
yurii-filiushkin: Ok. Thanks

TarasHoliuk (approved):
GJ :+1: 
Approved :heavy_check_mark:

PR #1530, opened 2023-02-26
:white_check_mark: implement solution for robotFactory kata

Bogdan-Mykhailov:
implement solution for robotFactory kata

### DIFF
--- a/src/robotFactory.js
+++ b/src/robotFactory.js
@@ -1,10 +1,81 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+  };
+
+  goBack(step = 1) {
+    this.coords.y -= step;
+  };
+
+  goRight(step = 1) {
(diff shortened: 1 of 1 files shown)


src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
+  };
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+
+  getInfo() {
+    return `Robot: ${this.name},`
+      + ' ' + `Chip version: ${this.chipVersion},`
+      + ' ' + `Weight: ${this.weight}`;
```
vladyslav-mosor, line 39:
```suggestion
    return `Robot: ${this.name}, `
      + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```

src/robotFactory.js
```diff
@@ -1,10 +1,104 @@
...
+  }
+
+  hookLoad(cargo) {
+    if (!cargo.weight || cargo.weight > this.maxLoadWeight) {
+      return;
+    }
+
+    if (!this.currentLoad) {
+      this.currentLoad = cargo;
+    }
```
vladyslav-mosor, line 95: Replace it with one if using && operator to check if cargo weight is less or equal to maxLoadWeight.

vladyslav-mosor (changes requested):

src/robotFactory.js
Bogdan-Mykhailov: 👍🏼

src/robotFactory.js
Bogdan-Mykhailov: 👍🏼

OleksiiBlezniuk (approved):

PR #1535, opened 2023-02-27
solution

Zelinskyi-Serhii:


src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+
+    if (cargo !== null) {
+      this.currentLoad.weight = cargo.weight;
+      this.currentLoad.description = cargo.description;
+    }
+
+    if (cargo === null || cargo.weight > this.maxLoadWeight) {
+      this.currentLoad = null;
+    }
+  }
```
gretoq, line 92: Too complicated solution for adding currentLoad where we have only two conditions

src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+    };
+  }
+
+  getInfo() {
+    const { name, chipVersion, weight } = this;
+
+    return `Robot: ${name}, Chip version: ${chipVersion}, Weight: ${weight}`;
+  }
+
+  goForward(x = 1) {
```
gretoq, line 23: Complicated understanding your code. 'x' is bad name for variables, especially where you have 'this.coords.y' and you add 'x' to it.

src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
 
-class DeliveryDrone {}
+    this.coords = {
+      x: x,
+      y: y,
+    };
+  }
+
+  getInfo() {
+    const { name, chipVersion, weight } = this;
```
gretoq, line 18: You can delete this row of code and add 'this' to 'name', 'chipVersion' and 'weight'

gretoq (changes requested):
Your implementation for 'hookLoad' is too complicated for understanding.

src/robotFactory.js
OleksiiBlezniuk: You may leave it as it is now, but it's strange to desctructure `this` object

src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
```
OleksiiBlezniuk, line 42: Btw, do `BaseRobot` constructor expect that there will be extra property in `coords` object?

src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+
+    if (currentLoad === undefined) {
+      this.currentLoad = null;
+    } else if (currentLoad === null) {
```
OleksiiBlezniuk, line 66: Just set `null` as default value to `currentLoad` param

src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
OleksiiBlezniuk, line 59: This line is too long, let's move each param to new line

src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+
+    this.maxLoadWeight = maxLoadWeight;
+
+    if (currentLoad === undefined) {
+      this.currentLoad = null;
+    } else if (currentLoad === null) {
+      this.currentLoad = {
+        weight: null,
+        description: null,
+      };
```
OleksiiBlezniuk, line 70: No, if there is no `currentLoad` passed - you need to set `null` to it, it should be either correct object or null

src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+    } else {
+      this.currentLoad = {
+        weight: currentLoad.weight,
+        description: currentLoad.description,
+      };
```
OleksiiBlezniuk, line 75: Just add `currentLoad` object, no need to re-create each property

src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+
+    if (cargo !== null) {
+      this.currentLoad.weight = cargo.weight;
+      this.currentLoad.description = cargo.description;
+    }
+
+    if (cargo === null || cargo.weight > this.maxLoadWeight) {
+      this.currentLoad = null;
+    }
+  }
```
OleksiiBlezniuk, line 92: Too complicated, please fix it
There are two options - if cargo is valid - set cargo to currentLoad, if not - do nothing

OleksiiBlezniuk (changes requested):

sergii-nosachenko (approved):

PR #1545, opened 2023-02-28
commit message

Peti51:


src/robotFactory.js
```diff
@@ -1,13 +1,83 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
polosanya, line 58:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion,
    maxLoadWeight, 
    currentLoad = null,
  ) {
```

src/robotFactory.js
```diff
@@ -1,13 +1,83 @@
...
+    super(name, weight, coords, chipVersion);
+
+    this.coords.z = coords.z || 0;
```
polosanya, line 61: No need to duplicate them again. You added all the coords in super

src/robotFactory.js
```diff
@@ -1,13 +1,83 @@
...
+  }
+}
 
 module.exports = {
   BaseRobot,
   FlyingRobot,
   DeliveryDrone,
 };
+
+// 0642510501
```
polosanya, line 83: What is it? Is it a phone number?)

polosanya (changes requested):
Almost perfect!

Peti51:
I tried to delete the line this.coords.z = coords.z || 0; from line 61 but it didn't worked.

Volodymyr-Melnyk-Fron10der (approved):
TextAnimatedGIF

PR #1546, opened 2023-03-01
Created a robot factory, which makes robots which have some abilities

gretoq:
When I did this solution. I learned:

- Class basic syntax;
- How to use inheritance in classes;
- How is super works.

### DIFF
--- a/src/robotFactory.js
+++ b/src/robotFactory.js
@@ -1,10 +1,129 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
 
-class FlyingRobot {}
+    {
+      x = 0,
+      y = 0,
+    } = {},
 
-class DeliveryDrone {}
+    chipVersion,
+  ) {
+    this.name = name;
+    this.weight = weight;
+
+    this.coords = {
+      x,
+      y,
+    };
+
+    this.chipVersion = chipVersion;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+      + `Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
+  }
+
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
+
+  goBack(step = 1) {
+    this.coords.y -= step;
+  }
+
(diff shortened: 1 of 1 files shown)


src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
 
-class FlyingRobot {}
+    this.name = name;
+    this.weight = weight;
 
-class DeliveryDrone {}
+    this.coords = {
+      x: x,
+      y: y,
+    };
```
illiarerun, line 13: hmm, can we write it in more short way? I mean, do we really need to repeat identifiers here?

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { z = 0 } = coords;
```
illiarerun, line 45: might be an overkill

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    currentLoad = null,
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    const weight = cargo.weight;
```
illiarerun, line 75: use destructuring instead of creating variable with the same name
```suggestion
    const { weight } = cargo;
```

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    const islessMaxLoad = weight <= this.maxLoadWeight;
```
illiarerun, line 76:
```suggestion
    const isLessMaxLoad = weight <= this.maxLoadWeight;
```

illiarerun (changes requested):

kuznetsova-anastasiia (approved):

sergii-nosachenko (approved):
Good work 👍

PR #1555, opened 2023-03-02
Solution

DenysDubas:


src/robotFactory.js
```diff
@@ -1,10 +1,100 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return (
+      `Robot: ${this.name}, Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`
+    );
+  }
```
MaksymMaliuk, line 41: I think in will be better in this way
```suggestion
 getInfo() {
    const { name, chipVersion, weight } = this;

    return `Robot: ${name}, Chip version: ${chipVersion}, Weight: ${weight}`;
  }
```

MaksymMaliuk (approved):
GJ

src/robotFactory.js
sergii-nosachenko: Good point. @DenysDubas you can refactor

sergii-nosachenko (approved):
Good work 👍
Read comments

PR #1577, opened 2023-03-05
Develop

yaroslav-vasilyev:


src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  /* eslint-disable */
```
YaroslavDzyadiv, line 33: You shouldn't be disabling linter for this task, in case you're using this command, it's better to disable it for one line.
```suggestion
  /* eslint-disable-next-line */
```

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
YaroslavDzyadiv, line 35: You can simply move part of your code to next line.
```suggestion
    return (
      `Robot: ${this.name}, Chip version: ${this.chipVersion}, `
      + `Weight: ${this.weight}`
    );
```

YaroslavDzyadiv (approved):
Awesome work!
It's better not to disable your linter, but you still have great code!

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0,
+  }, chipVersion) {
```
OleksiiBlezniuk, line 6: In this case - move each param to new line

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
OleksiiBlezniuk, line 12: But you already set 0 to these object params

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}`
+    + `, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
+  }, chipVersion) {
```
OleksiiBlezniuk, line 42: Function params should be written either in one line or every param should be at new line

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z || 0;
```
OleksiiBlezniuk, line 44: But 0 is already set to this param

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
+  }, chipVersion, maxLoadWeight, currentLoad = null) {
```
OleksiiBlezniuk, line 59: Move each param to new line

OleksiiBlezniuk (changes requested):

src/robotFactory.js
yaroslav-vasilyev: Yes. But its 0 for empty object, not for empty property 

image

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    + `, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, {
+    x = 0,
+    y = 0,
+    z = 0,
+  }, chipVersion) {
```
sergii-nosachenko, line 45:
```suggestion
  constructor(
    name, 
    weight, 
    {
      z = 0,
      ...baseCoords,
    }, 
    chipVersion,
  ) {
```
Better place each arg on a new line if there are more than 2 ones.
Also, you don't need x and y coords here - use rest operator.

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    super(name, weight, {
+      x,
+      y,
+    }, chipVersion);
```
sergii-nosachenko, line 49: Use baseCoords

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, {
+    x = 0,
+    y = 0,
+    z = 0,
+  }, chipVersion, maxLoadWeight, currentLoad = null) {
```
sergii-nosachenko, line 67: New lines.
You don't need to destructurize coords. It will be done inside `FlyingRobot`.

sergii-nosachenko (changes requested):

src/robotFactory.js
yaroslav-vasilyev: How i can use ...baseCoords at here?

image

src/robotFactory.js
```diff
@@ -1,10 +1,110 @@
...
+      ...baseCoords
+    },
+    chipVersion
+  ) {
+    super(
+      name,
+      weight,
+      {
+        ...baseCoords,
+      },
```
sergii-nosachenko, line 60:
```suggestion
      baseCoords,
```

src/robotFactory.js
```diff
@@ -1,10 +1,110 @@
...
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    {
+      ...baseCoords
+    },
```
sergii-nosachenko, line 81:
```suggestion
    coords,
```

src/robotFactory.js
```diff
@@ -1,10 +1,110 @@
...
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(
+      name,
+      weight,
+      {
+        ...baseCoords,
```
sergii-nosachenko, line 90: Why do you need to spread an object? Pass it as a parameter directly.

sergii-nosachenko (changes requested):

src/robotFactory.js
sergii-nosachenko: You don't need to create a new object here.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1581, opened 2023-03-05
solution

valeriiayarzhemska:


chernenko2 (approved):
All is good!

src/robotFactory.js
```diff
@@ -1,10 +1,120 @@
...
+    { x = 0, y = 0, z = 0 },
+    chipVersion
+  ) {
+    super(
+      name,
+      weight,
+      {
+        x,
+        y,
+        z,
```
OleksiiBlezniuk, line 63: do `BaseRobot` constructor expect that there will be 3 coords in this object?

OleksiiBlezniuk (approved):
GJ, but let's avoid passing unexpected arguments to functions

PR #1582, opened 2023-03-06
task completed

ArthurKrepko:


valeriiayarzhemska (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      'x': coords.x || 0,
+      'y': coords.y || 0,
+    };
```
OleksiiBlezniuk, line 11: Object keys are already strings
And let's use destructuring assignment for `coords`

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      ...coords,
+      'z': coords.z || 0,
+    };
```
OleksiiBlezniuk, line 45: `this` already has `coords` object, just set missing `z` coord

OleksiiBlezniuk: And keys are string in objects by deafult

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+
+    this.currentLoad = currentLoad || null;
+    this.maxLoadWeight = maxLoadWeight;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = {
+        ...cargo,
+      };
```
OleksiiBlezniuk, line 69: No need to create copy, it's enough in this case to add link to the object

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.currentLoad = currentLoad || null;
```
OleksiiBlezniuk, line 61: Assign `null` as default value

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
OleksiiBlezniuk, line 58: This line is too long, usually if we have more than 3 params in function we move each param to new line

OleksiiBlezniuk (changes requested):

src/robotFactory.js
illiarerun: again

src/robotFactory.js
illiarerun: and again

illiarerun (changes requested):

sergii-nosachenko (approved):

PR #1583, opened 2023-03-06
solution

milenadiamonds1:


src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+      y: y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this;
```
stanislavkulakovskyi, line 18: why you returning this in go methods?  :)

stanislavkulakovskyi (approved):
Good job! The only suggestion is to remove 'return this' in goForward, goBack, etc...

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, { x, y }, chipVersion) {
```
sergii-nosachenko, line 4:
```suggestion
  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
```
Set default values while destructuring

src/robotFactory.js
sergii-nosachenko, line 18: Why do you return this?

sergii-nosachenko (changes requested):

src/robotFactory.js
```diff
@@ -1,39 +1,31 @@
...
-  constructor(name, weight, { x, y }, chipVersion) {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
     this.name = name;
     this.weight = weight;
 
     this.coords = {
-      x: x || 0,
-      y: y || 0,
+      x: x,
+      y: y,
```
sergii-nosachenko, line 10: Checklist #1

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight,
+    currentLoad) {
```
sergii-nosachenko, line 56:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion, 
    maxLoadWeight,
    currentLoad,
  ) {
```

Also, set a default value for `currentLoad` here.

sergii-nosachenko (changes requested):

PR #1588, opened 2023-03-06
solution

vadym4che:


src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
sergii-nosachenko, line 4:
```suggestion
  constructor(
    name, 
    weight, 
    { x = 0, y = 0 }, 
    chipVersion,
  ) {
```
Better use a modern approach of setting default values using object restructuring.
Also, separate arguments with new lines if you have more than 3 ones.

src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = Object.assign({
+      x: 0, y: 0, z: 0,
+    }, coords);
```
sergii-nosachenko, line 39: You need only the `z` prop. Read the comment above.

src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
...
+  }
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null
```
sergii-nosachenko, line 53: Place each arg on new line

sergii-nosachenko (changes requested):

src/robotFactory.js
vadym4che: reduced

src/robotFactory.js
vadym4che: done

sergii-nosachenko (approved):

PR #1590, opened 2023-03-07
solution

Spec0s:


agniya-i (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    this.coords.x += steps;
+  }
+
+  goLeft(steps = 1) {
+    this.coords.x -= steps;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: \
+${this.chipVersion}, Weight: ${this.weight}`;
```
polosanya, line 34:
```suggestion
    return `Robot: ${this.name}, Chip version:`
      +`${this.chipVersion}, Weight: ${this.weight}`;
```
It is better to break it into two strings

src/robotFactory.js
```diff
@@ -1,10 +1,80 @@
...
+    this.coords.z += steps;
+  }
+
+  goDown(steps = 1) {
+    this.coords.z -= steps;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, position, chipVersion, maxLoadWeight, cargo) {
```
polosanya, line 59:
```suggestion
  constructor(
    name, 
    weight, 
    position, 
    chipVersion, 
    maxLoadWeight, 
    cargo,
  ) {
```

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comments)

PR #1599, opened 2023-03-09
created robot factory

JasterHim:


src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
+    this.coords = coords;
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
```
Viktor-Kost, line 10: BaseRobot doesn't have z-axis
```suggestion
```

Viktor-Kost, line 10: it should be as one instance
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

Viktor-Kost (changes requested):
Nice!

Need some improvements

src/robotFactory.js
JasterHim: Now I will add everything, but I have 1 logical question. BaseRobot does not have the ability to work with the z coordinate. This opportunity appears only for FlyingRobot and all who inherit from it. Why add a property to the BaseRobot class that it cannot use or change?

JasterHim: If I add the z coordinate here, the tests is feiled

AlisaVlasova (approved):

PR #1600, opened 2023-03-10
Solution

fisherok3452:


src/robotFactory.js
```diff
@@ -1,10 +1,117 @@
...
 
-class FlyingRobot {}
+    if (!coords.y) {
+      coords.y = 0;
+    }
 
-class DeliveryDrone {}
+    if (!coords.x) {
+      coords.x = 0;
+    }
```
tetiana-ivanchenko, line 25: if you use default values of args, do you need this code?

src/robotFactory.js
```diff
@@ -1,10 +1,117 @@
...
+  }
+
+  goForward(step = 1) {
+    return (
+      this.coords.y += step
+    );
+  };
```
tetiana-ivanchenko, line 32:
```suggestion
  goForward(step = 1) {
    this.coords.y += step;
    
    return this;
  };
```

src/robotFactory.js
```diff
@@ -1,10 +1,117 @@
...
+    coords = {
+      x: 0, y: 0, z: 0,
+    },
+    chipVersion
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    if (!coords.z) {
+      coords.z = 0;
+    }
```
tetiana-ivanchenko, line 70: the same

tetiana-ivanchenko (changes requested):

src/robotFactory.js
fisherok3452: test will not pass without this code

src/robotFactory.js
etojeDenys: still not fixed

src/robotFactory.js
etojeDenys: try to write like:
```
this.coords = {
  x: coords.x || 0;
  y: coords.y || 0;
}
```

etojeDenys (changes requested):

AlisaVlasova (approved):

PR #1602, opened 2023-03-10
Solution

iryna-serniuk:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.coords.y -= step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
```
etojeDenys, line 34: let's add a check so that x, y do not go negative

etojeDenys (changes requested):
almost done

etojeDenys (approved):
awesome

PR #1603, opened 2023-03-11
Classes done

oleksandr-plias:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    + `Chip version: ${this.chipVersion}, `
+    + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, { x, y, z = 0 }, chipVersion) {
+    super(name, weight, {
+      x, y,
+    }, chipVersion);
```
Volodymyr-Melnyk-Fron10der, line 41: you can do smth like
```suggestion
  constructor(name, weight, { z = 0, ...coords }, chipVersion) {
    super(name, weight, coords, chipVersion);
```

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
```
Volodymyr-Melnyk-Fron10der, line 56: I don't think you need this default, cause you set it in upper classes

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+
+      return true;
+    } else {
+      return false;
```
Volodymyr-Melnyk-Fron10der, line 69: Actually you don't have to return anything

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, good job! Check my comments plz

SanyaBratashchuk (approved):

PR #1604, opened 2023-03-11
solution :)

katerynashylina:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
Volodymyr-Melnyk-Fron10der, line 4: you can set defaults like this
```suggestion
  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
```
Volodymyr-Melnyk-Fron10der, line 11: and then just
```suggestion
    this.coords = { x, y };
```

Volodymyr-Melnyk-Fron10der (approved):
Good job! Just check one hint I left you for future

PR #1619, opened 2023-03-19
firstFactory

edkido:


src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
 
-class FlyingRobot {}
+    if (!coords.x) {
+      coords.x = 0;
+    }
 
-class DeliveryDrone {}
+    if (!coords.y) {
+      coords.y = 0;
+    }
```
OleziO, line 22: You don't need this code cause you initialize `x` and `y`.

OleziO (changes requested):

src/robotFactory.js
edkido: then it cannot pass 2 tests: should set coords to 0 by default (4ms)
    ✕ should set another coord to 0 by default if was passed only one (1ms)

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
etojeDenys, line 44: redundant, cause you set it in Baserobot

etojeDenys (changes requested):
almost done

etojeDenys (approved):
lgtm

PR #1621, opened 2023-03-20
robot factory done

2pasha:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  };
 
-class DeliveryDrone {}
+  getInfo() {
+    // eslint-disable-next-line max-len
```
VitaliyBondarenko1982, line 16: another option to split string on three parts and write them in constants

VitaliyBondarenko1982 (approved):

PR #1625, opened 2023-03-22
robotFactory

IvanFront-end:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z,
```
AlisaVlasova, line 11: z should not be in BaseRobot

AlisaVlasova (changes requested):

PR #1629, opened 2023-03-23
Solution

LBondarchuk:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    this.chipVersion = chipVersion;
+  }
+
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
+
+  goBack(step = 1) {
+    this.coords.y -= step;
+  }
```
etojeDenys, line 22: it is better to add a check so that our coordinates do not become negative

etojeDenys (changes requested):

polosanya (approved):
Good job!

PR #1635, opened 2023-03-27
Solution

BogdanParubchenk01:


etojeDenys (changes requested):
looks good, but let's add a check to make sure that our coordinates did not go negative

OleziO (approved):

anstsot (approved):
Great job, approved!

PR #1636, opened 2023-03-28
Solution

AnyaLyaLya:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = coords;
+
+    this.coords = {
+      x,
+      y,
+      z,
+    };
```
jstmpelowycz, line 51: do not reassign completely, just add what you need
```suggestion
    this.coords.z = coords.z || 0;
```

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
jstmpelowycz, line 65:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion,
    maxLoadWeight, 
    currentLoad = null
  ) {
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1638, opened 2023-03-28
Solution

igorkruz:


src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
...
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion,);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
```
PavloReutskiy, line 78:
```suggestion
    this.currentLoad = currentLoad || null;
```

image

src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
...
+      && this.currentLoad === null
+    ) {
+      this.currentLoad = cargo;
+    }
+  }
+
+  unhookLoad() {
+    if (this.currentLoad !== null) {
+      this.currentLoad = null;
+    }
```
PavloReutskiy, line 94: "if" is extra here, you just should set null to currentLoad in any cases
```suggestion
  unhookLoad() {
    this.currentLoad = null;
```
image

src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
...
+  ) {
+    super(name, weight, coords, chipVersion,);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (
+      cargo !== null
```
PavloReutskiy, line 83: I'm not sure, but maybe this line extra here
why cargo has to be null? even if it's an empty object it won't be null

src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+  constructor(name, weight, coords, chipVersion,) {
+    super(name, weight, coords, chipVersion,);
+
+    const { x = 0, y = 0, z = 0 } = coords;
+
+    this.coords = {
+      x,
+      y,
+      z,
+    };
```
jstmpelowycz, line 54: do not reassign it completely, just add what you need
```suggestion
    this.coords = coords.z || 0;
```

src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion,);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
```
jstmpelowycz, line 78: you do this step above
```suggestion
    this.currentLoad = currentLoad;
```

src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+  }
+
+  hookLoad(cargo) {
+    if (this.maxLoadWeight >= cargo.weight && this.currentLoad === null) {
```
jstmpelowycz, line 82:
```suggestion
    if (this.maxLoadWeight >= cargo.weight && !this.currentLoad) {
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1658, opened 2023-03-30
solution

Olekk17:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    // eslint-disable-next-line max-len
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
```
Nazarii98, line 32: You should add variables for every part of your string, don't use '// eslint-disable-next-line max-len'

Olekk17: Pasha said that it is acceptable to do that at such lines with long expresions in ` `

Nazarii98 (changes requested):
Great job!
Change one moment.

src/robotFactory.js
Nazarii98: Ok, you can change your work for the mentor review

Olekk17: No:
image
image
And this was accepted

Nazarii98: Good, I meant you can change the status of your work for the mentor review

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
```
SerhiiKirik, line 9: you can unite it
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0
   };
```

OR 

```suggestion
   this.coords = {
      x: 0,
      y: 0,
      ...coords,
    }
```

src/robotFactory.js
SerhiiKirik: Will be better if you avoid linter disabled

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && this.currentLoad === null) {
```
SerhiiKirik, line 65: You can just check like that
```suggestion
    if (cargo.weight <= this.maxLoadWeight && !this.currentLoad) {
```

SerhiiKirik (approved):
Well done 🚀

PR #1663, opened 2023-03-30
Solution

Vaskovics:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
Volodymyr-Melnyk-Fron10der, line 42: you have set x & y at the parent class already, so set only z here

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
```
Volodymyr-Melnyk-Fron10der, line 7:
```suggestion
  constructor(name, weight, { x = 0, y = 0 } , chipVersion) {
    this.name = name;
    this.weight = weight;

```

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
```
Volodymyr-Melnyk-Fron10der, line 11:
```suggestion
    this.coords = { x, y };
```

Volodymyr-Melnyk-Fron10der (changes requested):

src/robotFactory.js
Vaskovics: I cant change it like that as lint not allowed me

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x, y,
+    };
```
DarkMistyRoom, line 10:
```suggestion
    this.coords = {
      x, 
      y,
    };
```

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+    this.coords.x += step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, { x = 0, y = 0, z = 0 }, chipVersion) {
+    super(name, weight, {}, chipVersion);
```
DarkMistyRoom, line 37:
```suggestion
  constructor(name, weight, { x, y, z = 0 }, chipVersion) {
    super(name, weight, { x, y }, chipVersion);
```

DarkMistyRoom: 1. no need to set default values for x and y, it's already done in BaseRobot
2. why don't you pass them to the super? use existed functionality

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
DarkMistyRoom, line 55:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion,
    maxLoadWeight, 
    currentLoad = null
  ) {
```

DarkMistyRoom (changes requested):

src/robotFactory.js
Vaskovics: As I need to break the line and it didn't look readable enough.But I understood, will correct it. Thank you!

sTorba24 (approved):
It's cool!

PR #1666, opened 2023-03-30
add soltion

Pozdnya:


IvanTkachenko3301 (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = coords;
+
+    this.coords = {
+      x,
+      y,
+      z,
+    };
```
jstmpelowycz, line 52: do not reassign everything, just add `z`

jstmpelowycz (changes requested):

src/robotFactory.js
Pozdnya: done

src/robotFactory.js
```diff
@@ -1,10 +1,90 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
```
liliia-livandovskaya, line 69:
```suggestion
    super(
      name,
      weight,
      coords,
      chipVersion,
    );
```

src/robotFactory.js
```diff
@@ -1,10 +1,90 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    const robotName = `Robot: ${this.name}`;
+    const chipVer = `Chip version: ${this.chipVersion}`;
+    const weight = `Weight: ${this.weight}`;
+
+    return `${robotName}, ${chipVer}, ${weight}`;
```
liliia-livandovskaya, line 38:
```suggestion
    return `Robot: ${this.name}, Chip version: ${this.chipVersion},`
      +`Weight: ${this.weight}`;;
```
you can do something like that

liliia-livandovskaya (approved):

PR #1672, opened 2023-03-30
solution

Anastasia2403:


src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    const { x = 0, y = 0 } = coords;
 
-class FlyingRobot {}
+    this.name = name;
+    this.weight = weight;
 
-class DeliveryDrone {}
+    this.coords = {
+      x,
```
maksym-didyk, line 11: Maybe you can use x: coords.x || 0 (p. 4 checklist)

maksym-didyk (approved):
Good job!

src/robotFactory.js
Anastasia2403: p.1 checklist:  If in object variableName === keyName you can use short syntax.
I don`t know how would be better

SerhiiKirik: Up to you

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.8.10",
+    "@mate-academy/scripts": "^1.2.8",
```

SerhiiKirik (approved):
Great 🔥

PR #1676, opened 2023-03-31
create robofactory

ofenchuk:


yapelenua (approved):
Good Job)

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  };
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${
```
SerhiiKirik, line 32: better will be if you use destructuring here to avoid such wrap

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  goDown(step = 1) {
+    this.coords.z -= step;
+  };
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    coords = 0,
```
SerhiiKirik, line 61: It's extra
```suggestion
    coords,
```

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+  hookLoad(cargo) {
+    if ((cargo.weight <= this.maxLoadWeight)
+      && this.currentLoad === null) {
```
SerhiiKirik, line 72: you can just check like that
```suggestion
    if ((cargo.weight <= this.maxLoadWeight) && !this.currentLoad) {
```

SerhiiKirik (changes requested):
Well done 🔥 
Let's improve your code a bit 🧑‍💻

SerhiiKirik (approved):
Good job 🔥

PR #1683, opened 2023-04-01
solution

iHateAntichrist:


src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
```
Olekk17, line 44: Your parent constructor sets x and y values. You should set here only z 
image

Olekk17 (approved):
GJ

src/robotFactory.js
SerhiiKirik: Agree, fix it

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+    this.coords.x += step;
+  };
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  };
+
+  getInfo() {
+    // eslint-disable-next-line
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
SerhiiKirik, line 33: you can use destructuring here
```suggestion
    return `Robot: ${name}, Chip version: ${chipVersion}, Weight: ${weight}`;
```

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.8.10",
+    "@mate-academy/scripts": "^1.2.8",
```

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
```
SerhiiKirik, line 67:
```suggestion
  constructor(
      name,
      weight,
      coords,
      chipVersion,
      maxLoadWeight,
      currentLoad = null
    ) {
```

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻

PR #1684, opened 2023-04-01
robot factory solution

TolikSerhiychuk:


asmd564 (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    const robotsName = `Robot: ${this.name}`;
+    const robotChipVersion = `Chip version: ${this.chipVersion}`;
+    const robotName = `Weight: ${this.weight}`;
+
+    return `${robotsName}, ${robotChipVersion}, ${robotName}`;
```
liliia-livandovskaya, line 36:
```suggestion
    return `Robot: ${this.name}, Chip version: ${this.chipVersion},` 
      + `Weight: ${this.weight}`;
```
you can do something like that

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  // eslint-disable-next-line
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad = null) {
```
liliia-livandovskaya, line 57:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null,
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+    super(name, weight, coords, chipVersion);
```
liliia-livandovskaya, line 58:
```suggestion
    super(
      name,
      weight,
      coords,
      chipVersion
    );
```

liliia-livandovskaya (changes requested):
GJ! Lets work on fix

src/robotFactory.js
TolikSerhiychuk: got it

src/robotFactory.js
TolikSerhiychuk: fixed

src/robotFactory.js
TolikSerhiychuk: fixed

jstmpelowycz (approved):

PR #1691, opened 2023-04-04
solution

popeye88:


src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    { x = 0, y = 0 },
```
polosanya, line 7:
```suggestion
    { 
      x = 0, 
      y = 0,
    },
```

src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
+      + `Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
+  };
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(
+    name,
+    weight,
+    { z = 0, ...coords },
```
polosanya, line 47:
```suggestion
    coords,
```
It will work the same

polosanya (approved):
Good job!

PR #1694, opened 2023-04-06
Solution

meltourme:


package.json
```diff
@@ -16,14 +16,13 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.8.10",
+    "@mate-academy/scripts": "^1.2.8",
```

src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${
+      this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```
SerhiiKirik, line 38: you can use destructuring here

src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
+      coords,
+      chipVersion
+    );
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
SerhiiKirik, line 78: You can do just like that
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
+      this.currentLoad = cargo;
+    } else {
+      return this.currentLoad;
+    }
```
SerhiiKirik, line 82: also, you don't need this "else"
```suggestion
    }
    
      return this.currentLoad;
    }
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

SerhiiKirik (approved):
LGTM 🔥

PR #1697, opened 2023-04-08
solution

MikiOOOO:


src/robotFactory.js
```diff
@@ -1,10 +1,64 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords = {}, chipVersion, maxLoadWeight,
+    currentLoad = null) {
+    super(name, weight, chipVersion, coords);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
kasiakubisiak, line 54:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```
In this case this approach is more concise

src/robotFactory.js
```diff
@@ -1,10 +1,64 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+  getInfo() {
+    /* eslint-disable */
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+    /* eslint-enable */
```
kasiakubisiak, line 29: If it is possible, we should address eslint errors and fix them rather than disabling them.
```suggestion
    return `Robot: ${this.name}, `
    + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
```

kasiakubisiak (approved):

PR #1699, opened 2023-04-09
Fedenko Robot Factory

VladFedenko:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
 
-class DeliveryDrone {}
+    if (!this.coords.y) {
+      this.coords.y = 0;
+    }
+  }
+
+  goForward(y = 1) {
+    this.coords.y += y;
+  }
```
etojeDenys, line 21: let's add the check if `y` not negative

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #1702, opened 2023-04-10
add solution

allagutsul27:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
VitaliyBondarenko1982, line 62:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

VitaliyBondarenko1982 (approved):

src/robotFactory.js
allagutsul27: It will look better that way. Thanks for the comment and suggested changes!

PR #1705, opened 2023-04-12
robot factory was created

Daria13k:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+  };
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  };
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
TarasHoliuk, line 53:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
```
it's better to format it in this way (FYI)

TarasHoliuk (approved):
LGTM :+1:

PR #1709, opened 2023-04-13
Solution

MarinaOlenchenko:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, `
```
SerhiiKirik, line 33: you can make a destructuring from this here
```js
  const { name, ... } = this;
```

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    x: 0, y: 0, z: 0,
+  }, chipVersion, maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
SerhiiKirik, line 65: You can make it shortest
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

SanyaBratashchuk (approved):

PR #1712, opened 2023-04-13
solution

victor-buldenko:


src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
SerhiiKirik, line 72: you can make it shortest
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    const n = `Robot: ${this.name}`;
```
SerhiiKirik, line 32: what is n?
Like, n its Name, yeah?
You can make a destructuring from this here
```js
  const { 
  name, 
  chipVersion,
  weight,
 } = this;
```

SerhiiKirik (approved):
Well done 🚀

PR #1717, opened 2023-04-15
robotfactory

theFallcon:


src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
-class FlyingRobot {}
+    const {
+      x = 0,
+      y = 0,
+    } = coords;
 
-class DeliveryDrone {}
+    this.coords = {
+      x, y,
+    };
```
polosanya, line 15:
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```
You can set them right here without destructuring

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+    this.chipVersion = chipVersion;
+  }
+
+  getInfo() {
+    return [
+      `Robot: ${this.name}`,
+      `Chip version: ${this.chipVersion}`,
+      `Weight: ${this.weight}`,
+    ].join(', ');
```
polosanya, line 24: Just combine them with `+`, no need to overcomplicate

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  goUp(step = 1) {
+    this.coords.z += step;
+  }
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
polosanya, line 56:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion, 
    maxLoadWeight, 
    currentLoad,
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  }
+
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
+  goBack(step = 1) {
+    this.coords.y -= step;
+  }
+  goRight(step = 1) {
+    this.coords.x += step;
```
polosanya, line 34:
```suggestion
  goForward(step = 1) {
    this.coords.y += step;
  }
  
  goBack(step = 1) {
    this.coords.y -= step;
  }
  
  goRight(step = 1) {
    this.coords.x += step;
```
Add empty lines between methods to make your code easier to read

polosanya (changes requested):
Almost done!

src/robotFactory.js
Anastasiia-Svintsova: Not fixed. The array methods are expensive and can decrease productivity from time to time. So in this case it's better to use +
image

Anastasiia-Svintsova (approved):

PR #1718, opened 2023-04-16
task was done

anton-chornobai:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return (
+      `Robot: ${this.name},` +
+      `Chip version: ${this.chipVersion}, Weight: ${this.weight}`
+    );
```
VitaliyBondarenko1982, line 32:
```suggestion
    return (
      `Robot: ${this.name}, ` +
      `Chip version: ${this.chipVersion}, Weight: ${this.weight}`
    );
```
one test failed because no white space here
also you can make variables for each part of string to avoid long string

```
  const robot = `Robot: ${this.name}`;
  const chip = `Chip version: ${this.chipVersion}`;
  const weight = `Weight: ${this.weight}`;

  return `${robot}, ${chip}, ${weight}`;

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #1722, opened 2023-04-17
solution

kq333:


Amsko6 (approved):

src/robotFactory.js
```diff
@@ -1,16 +1,25 @@
...
+    },
+    chipVersion
+  ) {
     this.name = name;
     this.weight = weight;
 
-    const { x = 0, y = 0 } = coords;
+    if (coords.x === undefined) {
+      coords.x = 0;
+    }
```
kasiakubisiak, line 17:
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```
I'm not sure if checking for undefined for coords x and y is necessary here as though we assign default values in property assignment.

src/robotFactory.js
```diff
@@ -1,16 +1,25 @@
...
 
-    this.coords = {
-      x,
-      y,
-    };
+    if (coords.y === undefined) {
+      coords.y = 0;
+    }
```
kasiakubisiak, line 21:
```suggestion
```

kasiakubisiak (changes requested):

kasiakubisiak (approved):

PR #1726, opened 2023-04-17
First solution

marshalek37:


SuaBy514 (approved):

src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
```
kasiakubisiak, line 42:
```suggestion
  constructor(name, weight, coords,
```
You don't need to declare coords here - it already inherit the coords property from the BaseRobot

src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
```
kasiakubisiak, line 62: You don't need to declare coords here as well :)

src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
+  }, chipVersion, maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
kasiakubisiak, line 70:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

kasiakubisiak (changes requested):

PR #1729, opened 2023-04-17
solution

natalia-klonowska:


src/robotFactory.js
```diff
@@ -1,10 +1,69 @@
...
-class DeliveryDrone {}
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
kasiakubisiak, line 41:
```suggestion
    this.coords.z = coords.z || 0;
```
here you could just define the z coordinate because we don't need to do it for x and y as though they come from parent class - lets avoid the extra checks :)

src/robotFactory.js
```diff
@@ -1,10 +1,69 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+  getInfo() {
+    return 'Robot: ' + this.name
+    + ', Chip version: ' + this.chipVersion
+    + ', Weight: ' + this.weight;
```
kasiakubisiak, line 29:
```suggestion
    return `Robot: ${this.name}, Chip version: ${this.chipVersion},`
    + `Weight: ${this.weight}`;
```

kasiakubisiak (changes requested):

kasiakubisiak (approved):
nice!

PR #1731, opened 2023-04-18
first solution

Ozwiena:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
kasiakubisiak, line 42:
```suggestion
    this.coords.z = coords.z || 0;
```
here you could just define the z coordinate because we don't need to do it for x and y as though they come from parent class - lets avoid the extra checks :)

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad !== null) {
+      return;
+    }
+
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
kasiakubisiak, line 65:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

kasiakubisiak (changes requested):

PR #1736, opened 2023-04-19
Develop

HannaVasylieva:


src/robotFactory.js
```diff
@@ -1,10 +1,116 @@
...
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
```
mykhalenych, line 17:
```suggestion
  goForward(step = 1) {
    if (step > 0) {
```
add to all methods

src/robotFactory.js
```diff
@@ -1,10 +1,116 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return 'Robot: '
```
mykhalenych, line 34: better to use template string

src/robotFactory.js
```diff
@@ -1,10 +1,116 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
+  }, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z || 0;
+  }
+
+  goBack(step) {
```
mykhalenych, line 48: you don't need to add methods from BaseRobot you already inherited them

src/robotFactory.js
```diff
@@ -1,10 +1,116 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+  goBack(step) {
```
mykhalenych, line 80: the same

mykhalenych (changes requested):

src/robotFactory.js
HannaVasylieva: I know, but if I change something in that string my test is failed. I asked question in chat and we find that solution

mykhalenych (approved):

PR #1744, opened 2023-04-23
solution

yuron-maker:
I know that the output of information is written incorrectly, but I was killed by the error of too long a line when everything is in one return

### DIFF
--- a/src/robotFactory.js
+++ b/src/robotFactory.js
@@ -1,10 +1,86 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: x,
+      y: y,
+    };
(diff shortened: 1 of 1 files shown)


src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    let info = `Robot: ${this.name}, Chip version: ${this.chipVersion},`;
+
+    info += ` Weight: ${this.weight}`;
+
+    return info;
```
tetiana-ivanchenko, line 36: if you have any questions or know, that you do incorrect, you shouldn't create the PR, you should ask your question in the fe_chat
```suggestion
   return `Robot: ${this.name}, Chip version: ${this.chipVersion},`
     + ` Weight: ${this.weight}`;
```

tetiana-ivanchenko (changes requested):

src/robotFactory.js
yuron-maker: Thanks)

OleziO (approved):

PR #1749, opened 2023-04-25
Solution_1

alexxxstep:


src/robotFactory.js
```diff
@@ -1,10 +1,118 @@
...
+      y: 0,
+    };
+
+    if (Object.keys(coords).length !== 0) {
+      this.coords = {
+        ...this.coords,
+        ...coords,
+      };
+    }
+  }
```
DarkMistyRoom, line 33: This approach is too complicated. Would be enough to decompose `coords `and give them default values

DarkMistyRoom (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.unhookLoad();
+    this.hookLoad(currentLoad);
+  }
+
+  hookLoad(cargo) {
+    if (cargo) {
+      if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
+        this.currentLoad = cargo;
```
anstsot, line 75:
```suggestion
    if (cargo && !this.currentLoad && cargo.weight <= this.maxLoadWeight) {
      this.currentLoad = cargo;
```

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+    this.coords.x += step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    const a = `Robot: ${this.name}`;
+    const b = `Chip version: ${this.chipVersion}`;
+    const c = `Weight: ${this.weight}`;
```
anstsot, line 37: better to give more meaningful names

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+    this.coords = {
+      x: 0,
+      y: 0,
+    };
 
-class DeliveryDrone {}
+    this.coords = {
+      ...this.coords,
+      ...coords,
+    };
```
anstsot, line 16: please decompose it and set `this.coords` with default x and y value by using `||`

anstsot (changes requested):
It’s almost done! Let’s improve a little

anstsot (approved):
Great job, approved!

PR #1751, opened 2023-04-26
initial commit

44shaggy:


src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`; // eslint-disable-line
```
Anastasiia-Svintsova, line 32: It's not a good practice to disable eslint. Would be better to refactor it, to avoid linter errors

src/robotFactory.js
```diff
@@ -1,10 +1,70 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
Anastasiia-Svintsova, line 52:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion, 
    maxLoadWeight, 
    currentLoad
  ) {
```

Anastasiia-Svintsova (approved):

PR #1757, opened 2023-04-30
solution

zvir91:


src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = coords;
+
+    this.coords = {
```
Anastasiia-Svintsova, line 46: Instead of reassigning all coordinates in FlyingRobot's constructor, you can just add the z coordinate to the existing coords object: 'this.coords.z = z;'.

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+    return 'Robot: '
+    + `${this.name}, Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = coords;
```
Anastasiia-Svintsova, line 44: It is not necessary to redefine x and y coordinates in FlyingRobot's constructor as they are already set in the BaseRobot constructor.

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords,
+    chipVersion, maxLoadWeight, currentLoad = null) {
```
Anastasiia-Svintsova, line 64:
```suggestion
  constructor(
    name, 
    weight, 
    coords,
    chipVersion, 
    maxLoadWeight, 
    currentLoad = null
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && this.currentLoad === null) {
```
Anastasiia-Svintsova, line 71:
```suggestion
    if (cargo.weight <= this.maxLoadWeight && !this.currentLoad) {
```

Anastasiia-Svintsova (changes requested):

polosanya (approved):
Nice!

PR #1766, opened 2023-05-04
solution

antonina-klishch:


src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
 
-class FlyingRobot {}
+    this.name = name;
+    this.weight = weight;
 
-class DeliveryDrone {}
+    this.coords = {
+      x: x,
+      y: y,
+    };
```
IvanFesenko, line 13:
```suggestion
    this.coords = {
      x,
      y,
    };
```

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x = 0, y = 0, z = 0 } = coords;
+
+    this.coords = {
+      x: x,
+      y: y,
+      z: z,
```
IvanFesenko, line 49: x, y is redudant here, they will set is main class
```suggestion
    const { z = 0 } = coords;

    this.coords = {
      z,
```

IvanFesenko (approved):
Small remarks are in comments

PR #1770, opened 2023-05-06
Solution

AV1788:


src/robotFactory.js
```diff
@@ -1,10 +1,112 @@
...
 
-class FlyingRobot {}
+    const { x, y } = coords;
 
-class DeliveryDrone {}
+    if (Object.keys(coords).length === 0) {
+      this.coords = {
+        x: 0, y: 0,
+      };
+    } else if (y === undefined) {
```
TarasHoliuk, line 14: what if `x === undefined`?) instead of checking each case just set the default value for coords, x, and y in arguments/destructuring assignment

src/robotFactory.js
```diff
@@ -1,10 +1,112 @@
...
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+
+    if (currentLoad === undefined) {
```
TarasHoliuk, line 89: set default value in arguments

src/robotFactory.js
```diff
@@ -1,10 +1,112 @@
...
+      this.currentLoad = null;
+    }
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === undefined && cargo.weight <= this.maxLoadWeight) {
```
TarasHoliuk, line 95:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```
it's shorter and covers cases with `null` and `undefined`

src/robotFactory.js
```diff
@@ -1,10 +1,112 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x, y, z } = coords;
+
+    if (z === undefined) {
```
TarasHoliuk, line 60: the same as above

TarasHoliuk (changes requested):

OleziO (approved):

PR #1777, opened 2023-05-09
Solution

yurii-shkrobut-m:


src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion) {
```
DarkMistyRoom, line 8:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion
  ) {
```

DarkMistyRoom: fix it for all constructors

DarkMistyRoom (approved):
Good job!

PR #1780, opened 2023-05-11
solution

generatedman:


src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion) {
+    const { x, y } = coords;
```
dinaminahi, line 9: X and Y can be given default values like: 

`
    const { x = 0, y = 0 } = coords;

`

and this way, you won`t need x || 0

src/robotFactory.js
```diff
@@ -1,10 +1,91 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { x, y, z } = coords;
```
dinaminahi, line 50: same here, 
but x and y were set in the base class, so you can skip them and set only z

dinaminahi (approved):
Looks good 💪 

But check few comments

PR #1792, opened 2023-05-16
1st commit

ihor-halkin:


src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      ...coords,
+    };
```
polosanya, line 12:
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```
It's actually redundant, as you have only x and y coords

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return [
+      'Robot: ' + this.name,
+      'Chip version: ' + this.chipVersion,
+      'Weight: ' + this.weight,
+    ].join(', ');
```
polosanya, line 38: Just use template strings, combined with `+`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      ...this.coords,
+      z: coords.z || 0,
+    };
```
polosanya, line 49:
```suggestion
    this.coords.z = coords.z || 0;
```

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
```
polosanya, line 64:
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion, 
    maxLoadWeight, 
    currentLoad,
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,86 @@
...
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (!this.currentLoad) {
+      this.currentLoad = cargo;
+    }
+
+    if (cargo.weight > this.maxLoadWeight) {
+      this.currentLoad = null;
```
polosanya, line 76: 'hookLoad' method shoudn't set null as currentLoad. Just add this condition to the check above

polosanya (changes requested):
Good job! Let's improve your code a bit)

agniya-i (approved):

PR #1793, opened 2023-05-17
solution

skinnyjazz:


src/robotFactory.js
```diff
@@ -1,10 +1,101 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    { x = 0, y = 0 },
```
33Orange, line 7: Lets change this to `coords` and the default values ​​will be inserted already in the constructor

src/robotFactory.js
```diff
@@ -1,10 +1,101 @@
...
+  ) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x,
+      y,
+    };
```
33Orange, line 17: This will make it more readable and understandable.

```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

33Orange (changes requested):
Well done ! 🔥 

A few moment with readability

silchyn (approved):

PR #1805, opened 2023-05-24
Solution

julshvets:


src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+  };
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${
+      this.chipVersion}, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, { x = 0, y = 0, z = 0 }, chipversion) {
```
dinaminahi, line 39: There is no need to redefine 'x' and 'y' in FlyingRobot constructor, as it is already defined in the BaseRobot constructor. Instead,  write `coords`  and just add the z coordinate to 'this.coords'.

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, { x = 0, y = 0, z = 0 }, chipversion,
```
dinaminahi, line 57: same here

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+      x,
+      y,
+      z,
+    }, chipversion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
dinaminahi, line 70:
```suggestion
    if (!this.currentLoad  && cargo.weight <= this.maxLoadWeight) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+      this.currentLoad = { ...cargo };
```
dinaminahi, line 71: I don't think you need a new object here

dinaminahi (changes requested):
Good job! 👍 
Few moments to improve it

src/robotFactory.js
julshvets: Done

src/robotFactory.js
julshvets: Done

src/robotFactory.js
julshvets: I tried to write "coords" instead of {x, y, z=0}, but the tests didn't pass. According to tests the third constructor argument should be the coordinates of the robot, including z, and set z coordinate to 0 by default.  And the fourth constructor argument has to be a chipVersion. When I wrote just coords in arguments, there was no z coordinate inside it and tests didn't pass.

dinaminahi: yes, for default setting you could use || operator:
`this.coords.z = coords.z || 0`

dinaminahi (approved):
Well done 👍

src/robotFactory.js
julshvets: Thank you! You are right! Tests passed. I didn't use || operator...

PR #1807, opened 2023-05-25
solution

okosohor:


src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
+    this.chipVersion = chipVersion;
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this;
```
33Orange, line 17: `return this` here is not needed. Look carefully at the task

`goForward, goBack, goRight and goLeft methods take a step argument (1 by default) and move the robot by step in the appropriate direction`

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+  }
+
+  goBack(step = 1) {
+    this.coords.y -= step;
+
+    return this;
```
33Orange, line 23: Same as in `goForward`

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+
+    return this;
```
33Orange, line 29: Same as in `goForward`

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+
+    return this;
```
33Orange, line 35: Same as in previous methods

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+    return `Robot: ${this.name}, Chip version: \
+${this.chipVersion}, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z || 0;
+  }
```
33Orange, line 48: You can use rest and spread operators here.

```suggestion
  constructor(...args) {
    super(...args);
    this.coords.z = coords.z || 0;
  }
```

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+
+  goUp(step = 1) {
+    this.coords.z += step;
+
+    return this;
```
33Orange, line 53: Same as in previous movement methods

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+
+    return this;
```
33Orange, line 59: Same as in previous movement methods

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion, maxLoadWeight);
```
33Orange, line 66: None of the parent classes accept maxLoadWeight as parameters. Why pass it through super() ?

```suggestion
    super(name, weight, coords, chipVersion);
```

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
 };
+
+// const cargo1 = {
+//   weight: 57,
+//   description: 'cargo 57 weight',
+// };
+
+// const robot = new DeliveryDrone('X', 9, {}, 0.7, 78, cargo1);
+
+// console.log(robot);
```
33Orange, line 95: Remove redundant comments

src/robotFactory.js
```diff
@@ -1,13 +1,95 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion, maxLoadWeight);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad == null && cargo.weight <= this.maxLoadWeight) {
```
33Orange, line 72:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

33Orange (changes requested):
Well done !

A few moment to fix

src/robotFactory.js
okosohor: I dont understand how will we get coords? coords.z will be is not defined if we will use ...arg.

src/robotFactory.js
okosohor: I know there is no such check in the tests, but I thought it would be appropriate to add the ability to move several times as example : robot.goBack().goLeft(5).goForward()

src/robotFactory.js
okosohor: class FlyingRobot extends BaseRobot {
  constructor(...args) {
    super(...args);

  const [name, weight, coords, chipVersion] = args;

  }
 i can make something like this, but i think it`l be worse version than now and linter gives errors because we dont use name, weight, chipVersion.

33Orange (approved):
Well done ! 🔥

PR #1810, opened 2023-05-25
add task solution

KatyaZaremska:


src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+       + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
```
33Orange, line 39: You can use rest and spread operators in such cases
```suggestion
  constructor(...args) {
    super(...args);
```

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
33Orange, line 60:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

33Orange (changes requested):
Well done !

A few moments

src/robotFactory.js
KatyaZaremska: It doesn't work when I use operator

33Orange (approved):
Well done!

PR #1813, opened 2023-05-26
solution

Yuliya1907:


src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    const { x = 0, y = 0 } = coords;
```
naziamoff, line 8: why not just assign it with dot notation? I believe there's no reason to destructure it here

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    // eslint-disable-next-line max-len
```
naziamoff, line 34: nope, let's fix linting errors instead of disabling them

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords = {},
+    chipVersion, maxLoadWeight, currentLoad = null) {
```
naziamoff, line 63: if you move one prop to the new line - you move each of the props to the new line

naziamoff (changes requested):
a couple of important fixes and good to go 🚀

PR #1814, opened 2023-05-26
add functionality to robotFactory

cephy77:


src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
+    this.coords = coords;
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
```
naziamoff, line 10:
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    }
```

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+  }
 
-class FlyingRobot {}
+  getInfo() {
+    return (
+      `Robot: ${this.name},`
+      + ` Chip version: ${this.chipVersion}, Weight: ${this.weight}`
+    );
```
naziamoff, line 17: inconsistency kills readability. Let's have it concatenated instead

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
```
naziamoff, line 59:
```suggestion
    currentLoad = null
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad !== null) {
```
naziamoff, line 66: what if the load is undefined?

naziamoff (changes requested):
not bad 👍 

There are a couple of issues with readability though. Let's fix it and we're good

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+    }
+
+    return `Maximum capacity of this drone is ${this.maxLoadWeight}.
+    The cargo exceeds capacity by ${cargo.weight - this.maxLoadWeight} `;
+  }
+
+  unhookLoad() {
+    if (this.currentLoad === null) {
+      return 'The cargo compartment is already empty';
+    }
```
naziamoff, line 87: better to check for falseness instead
```suggestion
    if (!this.currentLoad) {
      return 'The cargo compartment is already empty';
    }
```

naziamoff (approved):
good job!

src/robotFactory.js
cephy77: Oh, miss that one

PR #1818, opened 2023-05-27
Solution

andrijdudar:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+    this.coords.x += step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, \
+Chip version: ${this.chipVersion}, \
+Weight: ${this.weight}`;
```
ivbutsykin, line 31: Ha, interesting decision. You can also use `+`, like:
```
    return (
      `Robot: ${this.name}, Chip version:`
      + ` ${this.chipVersion}, Weight: ${this.weight}`
    );
```

ivbutsykin (approved):
Good job!

PR #1819, opened 2023-05-28
add solution

niaYaro:


vladyslav-lapshyn (approved):
Good Job 🔥

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+  }
+
+  goUp(step = 1) {
+    this.coords.z += step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords,
+    chipVersion, maxLoadWeight, currentLoad = null) {
```
naziamoff, line 59: if you move one prop to the new line - you move each of them to the new line
```suggestion
  constructor(
    name, 
    weight, 
    coords,
    chipVersion, 
    maxLoadWeight, 
    currentLoad = null
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 67: is there any specific reason to check for null explicitly? Let's check for the falseness instead

naziamoff (approved):
not bad! But there are still a couple of suggestions

src/robotFactory.js
niaYaro: Ok, noted. thx!

PR #1827, opened 2023-05-29
Solution

Shevtsooov:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${
+      this.chipVersion}, Weight: ${this.weight}`;
+  }
```
naziamoff, line 34: let's add some consistency here

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 63: I guess it would be better to have it separated in the variable

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
naziamoff, line 55: consistency in your code - is a key for its readability. You can't just move one or two properties not moving the others
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion,
    maxLoadWeight, 
    currentLoad = null
  ) {
```

naziamoff (approved):
not bad, but there are some suggestions

PR #1829, opened 2023-05-29
solution

kirilshorin:


src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
```
naziamoff, line 4: it is better not to destructure here, as the code loses its readability

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x,
+      y,
+    };
```
naziamoff, line 12: and this way you can have assign the existing coords here
```suggestion
    this.coords = coords;
```

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null
+      && cargo.weight <= this.maxLoadWeight
```
naziamoff, line 71: better to create a variable for that

naziamoff (approved):
not bad 👍 
Some suggestions though

PR #1830, opened 2023-05-29
Solution

bohdan-mykhailenko:


src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+    name,
+    weight,
+    coords,
+    chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.hasOwnProperty('x') ? coords.x : 0,
+      y: coords.hasOwnProperty('y') ? coords.y : 0,
+      z: coords.hasOwnProperty('z') ? coords.z : 0,
```
naziamoff, line 54: same here

src/robotFactory.js
```diff
@@ -1,10 +1,93 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
```
naziamoff, line 74:
```suggestion
    currentLoad = null
  ) {
```

naziamoff (changes requested):
not bad, but some polishing required

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+    coords,
+    chipVersion,
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
33Orange, line 57: You already check `x` and `y` inside `BaseRobot`. 
Why do you need another check? 

```suggestion
    this.coords.z = coords.z || 0;
```

src/robotFactory.js
```diff
@@ -1,10 +1,96 @@
...
+    currentLoad = null,
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && this.currentLoad === null) {
```
33Orange, line 85: If `this.currentLoad` will be `undefined` ? 

33Orange (changes requested):
Almost done !

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+    currentLoad = null,
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && !this.currentLoad) {
```
33Orange, line 81: First of all let's check if your drone already have a cargo.

```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

33Orange (approved):
Good job

naziamoff (approved):

PR #1831, opened 2023-05-29
3 robot classes implemented

olehmahera:


src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+  }
+
+  goLeft(x = 1) {
+    this.coords.x -= x;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+    + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
```
naziamoff, line 34: let's keep it consistent, looks quite messy now

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+
+  goDown(z = 1) {
+    this.coords.z -= z;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
naziamoff, line 54: same as we used to do in HTML with the attributes. If you move one property to a new line - you move the others as well
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion,
    maxLoadWeight, 
    currentLoad = null
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,73 @@
...
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+  hookLoad(cargo) {
+    this.currentLoad = this.currentLoad === null
+    && cargo.weight <= this.maxLoadWeight
+      ? cargo
+      : this.currentLoad;
+  }
```
naziamoff, line 65: looks messy. Let's find a way to improve its readability

naziamoff (changes requested):
not bad, but there are some suggestions

src/robotFactory.js
olehmahera: It's length is more than 80, maybe, I'll just split it into 3 lines?

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+  hookLoad(cargo) {
+    if (this.currentLoad === null
```
33Orange, line 67: What happens if `this.currentLoad === undefined` ?

33Orange (changes requested):
Almost done !

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null) {
```
naziamoff, line 60:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+  hookLoad(cargo) {
+    if (!this.currentLoad
+    && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 68: we use this approach if we want to move the if condition to the new line
```suggestion
    if (
      !this.currentLoad
      && cargo.weight <= this.maxLoadWeight
    ) {
```

naziamoff (approved):
good job! A couple of extra minor suggestions:

PR #1832, opened 2023-05-29
add solution

ulositska:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
naziamoff, line 55: semi-pregnant code 😄 
It can't be this way. If you move one property to the new line - **each of them** should be moved to a new line
Consistency is the key
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion,
    maxLoadWeight, 
    currentLoad = null
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 63: what if the current load is undefined? Let's check for the falseness instead

naziamoff (changes requested):
not bad, but consider my suggestions in your following pull requests

src/robotFactory.js
ulositska: Yes, I was working on making it fully-pregnant, but got distracted 😬

src/robotFactory.js
ulositska: Do you mean if the 'undefined' will get directly passed to it? I thought that null will be enough, since I set it as a default value:thinking:

ulositska: Oh actually, if I check for falseness instead, it will help in case a zero is passed, nvm 🙈

33Orange (approved):
Well done !

naziamoff (approved):

PR #1836, opened 2023-05-29
Solution

ivankhrsh:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, Weight: ${
+      this.weight}`;
```
naziamoff, line 33: looks messy, let's think of something which would make it look more consistent

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords.z = !(coords.hasOwnProperty('z'))
+      ? 0
+      : coords.z;
```
naziamoff, line 43: overkill
```suggestion
    this.coords.z = coords.z || 0
```

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight,
+    currentLoad = null) {
```
naziamoff, line 57: if you move one prop to the new line - you move each of them to the new line. Consistency is the key
```suggestion
  constructor(
    name, 
    weight, 
    coords, 
    chipVersion, 
    maxLoadWeight,
    currentLoad = null
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    super(name, weight, coords, chipVersion);
+
+    this.currentLoad = currentLoad;
+    this.maxLoadWeight = maxLoadWeight;
+  }
+
+  hookLoad(cargo) {
+    const canHookCargo = this.currentLoad === null
```
naziamoff, line 65: what if the load is undefined? Let's check for the falseness instead

naziamoff (changes requested):
not bad but there are a couple of severe code style issues

src/robotFactory.js
naziamoff: not fixed

naziamoff (approved):
good job! But be more attentive next time and fix all the suggestions pls

PR #1838, opened 2023-05-29
add robotFactory task solution

opiquem:


src/robotFactory.js
```diff
@@ -1,10 +1,102 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    {
+      x = 0,
+      y = 0,
+    },
```
naziamoff, line 10: that decreases readability really hard. Let's have it just as coords

src/robotFactory.js
```diff
@@ -1,10 +1,102 @@
...
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && this.maxLoadWeight >= cargo.weight) {
```
naziamoff, line 91: better to check for the falseness instead
```suggestion
    if (!this.currentLoad&& this.maxLoadWeight >= cargo.weight) {
```

naziamoff (approved):
GJ 👍

PR #1840, opened 2023-05-29
made task js robot factory

vladyslav-lapshyn:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, `
+    + `Weight: ${this.weight}`;
```
naziamoff, line 33: let's concatenate each part, looks inconsistent now

naziamoff (approved):
well done 🔥

PR #1841, opened 2023-05-29
solution

vitvadym:


src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name, weight, { x = 0, y = 0 },
+    chipVersion) {
```
naziamoff, line 6: - if you move one property to the new line - the other properties should take one line **each**
- not really a good idea to destructure coords here. If you write it explicitly like `coords.x` it will not take a lot of space, but this way you significantly decrease readability

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${
+      this.chipVersion}, Weight: ${this.weight}`;
```
naziamoff, line 34: inconsistent, let's concatenate instead

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null
+  ) {
```
naziamoff, line 59: see my first comment

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 66: what if currentload is undefined, not null? Let's check for falseness instead

naziamoff (changes requested):

silchyn (approved):

naziamoff (approved):

PR #1842, opened 2023-05-29
solutin

snuber-w:


src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVer) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVer;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x ? coords.x : 0,
+      y: coords.y ? coords.y : 0,
```
naziamoff, line 11: - ternaries are always multilined
- you can achieve the same result with logical operator `||`

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${
+      this.chipVersion}, Weight: ${this.weight
+    }`;
```
naziamoff, line 34: inconsistent, let's concatenate instead

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVer) {
+    super(name, weight, coords, chipVer);
+
+    this.coords = {
+      x: coords.x ? coords.x : 0,
+      y: coords.y ? coords.y : 0,
+      z: coords.z ? coords.z : 0,
```
naziamoff, line 45: see my prev comment

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVer,
+    maxLoadWeight, currentLoad = null) {
```
naziamoff, line 60: inconsistent, if you move one argument to the new line - you move each of them to the new line

src/robotFactory.js
```diff
@@ -1,10 +1,79 @@
...
+    super(name, weight, coords, chipVer);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 68: what if the load is undefined, but not null?

src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
+  }
+}
```
33Orange, line 35:
```suggestion
  getInfo() {
    return `Robot: ${this.name}, `
    +`Chip version: ${this.chipVersion}, `
    + `Weight: ${this.weight}`;
  }
}
```

src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
+    weight,
+    coords,
+    chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
33Orange, line 48: You already check `x` and `y` inside BaseRobot.
Why do you need another check?
```suggestion
    this.coords.z = coords.z || 0;
```

src/robotFactory.js
```diff
@@ -1,10 +1,85 @@
...
+    maxLoadWeight,
+    currentLoad) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
33Orange, line 74: Check @naziamoff comment.
If `this.currentLoad` will be `undefined` ?

33Orange (changes requested):
Almost done

src/robotFactory.js
snuber-w: I'm not completely sure, but I think this check can solve it `this.currentLoad = cargo || null;`

33Orange: @snuber-w just change you condition to avoid such situations

`!this.currentLoad` such a check will solve all possible cases

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+    + `Chip version: ${this.chipVersion}, `
+    + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name,
+    weight,
+    coords,
+    chipVersion) {
```
naziamoff, line 42:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,82 @@
...
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad) {
```
naziamoff, line 63:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad
  ) {
```

src/robotFactory.js
naziamoff: not fixed

naziamoff (changes requested):

src/robotFactory.js
snuber-w: This confuses me a bit because the tests pass in all cases. I will try another check

33Orange (approved):
Well done !

naziamoff (approved):

PR #1843, opened 2023-05-29
add solution

serentinos:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {};
+    this.coords.x = coords.x || 0;
```
naziamoff, line 10: let's declare object properties inside the object instead of the dot notation

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+};
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, coords = {}, chipVersion, maxLoadWeight, currentLoad
+  ) {
```
naziamoff, line 56: I would move each prop to the new line

naziamoff (approved):
well done! A couple of small suggestions

PR #1844, opened 2023-05-29
Solution

MaksimKad:


src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
```
naziamoff, line 9: let's create a single coords object and assign its properties instead of creating both object and properties

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, \
+Chip version: ${this.chipVersion}, \
+Weight: ${this.weight}`;
+  }
```
naziamoff, line 33: something looks wrong here 👀 
Let's use concatenation instead

src/robotFactory.js
```diff
@@ -1,10 +1,71 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad || null;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
naziamoff, line 60:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

naziamoff (changes requested):
fix a couple of code style issues and we're good

dinaminahi (approved):
Well done 👍

naziamoff (approved):

PR #1848, opened 2023-05-29
Solution

PanyaPrimal:


src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    position = {
+      x: 0,
+      y: 0,
+    },
+    chipVersion,
+  ) {
```
33Orange, line 12: You will check the presence of properties in the constructor.
You are assigning a new object as a default value, and if you are given an empty object in this argument, this check will not work

```suggestion
  constructor(
    name,
    weight,
    position,
    chipVersion,
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+  constructor(
+    name,
+    weight,
+    position = {
+      x: 0,
+      y: 0,
+      z: 0,
+    },
+    chipVersion,
+  ) {
```
33Orange, line 56: Same as in `BaseRobot`

```suggestion
  constructor(
    name,
    weight,
    position,
    chipVersion,
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+  constructor(
+    name,
+    weight,
+    position = {
+      x: 0, y: 0, z: 0,
+    },
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null,
+  ) {
```
33Orange, line 80: Same as in prev robots
```suggestion
  constructor(
    name,
    weight,
    position,
    chipVersion,
    maxLoadWeight,
    currentLoad = null,
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,98 @@
...
+    super(name, weight, position, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
33Orange, line 87: If `this.currentLoad` will be `undefined` ? 

33Orange (changes requested):
Good job. 

A few moments to fix

silchyn (approved):

PR #1866, opened 2023-06-05
solution

Vo7kov:


src/robotFactory.js
```diff
@@ -1,10 +1,101 @@
...
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion
+  ) {
+    this.name = name;
+    this.weight = weight;
+    this.weight = weight;
```
IvanFesenko, line 13: looks like same data
```suggestion
    this.weight = weight;

```

src/robotFactory.js
```diff
@@ -1,10 +1,101 @@
...
+    name,
+    weight,
+    coords,
+    chipVersion
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
IvanFesenko, line 54: No need to set them again? They will be added in BaseRobot constructor
```suggestion

```

src/robotFactory.js
```diff
@@ -1,10 +1,101 @@
...
+    maxLoadWeight,
+    currentLoad
+  ) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
IvanFesenko, line 83: Same here, will be used FlyingRobot constructor
```suggestion

```

IvanFesenko (changes requested):
GJ. Check my commments

IvanFesenko (approved):

PR #1869, opened 2023-06-06
Solution

artyomwhite:


src/robotFactory.js
```diff
@@ -1,10 +1,90 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = {
+    x: 0, y: 0, z: 0,
+  }, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    if (coords.z === undefined) {
```
etojeDenys, line 53: Remove unnecessary conditional statements for setting coords.z, since it is already initialized in the constructor's default parameters.

etojeDenys (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (cargo.weight <= this.maxLoadWeight && this.currentLoad === null) {
```
VitaliyBondarenko1982, line 61:
```suggestion
    if (cargo.weight <= this.maxLoadWeight && !this.currentLoad) {
```

VitaliyBondarenko1982 (approved):
Looks good

PR #1875, opened 2023-06-07
Solution

RodionSav:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
vlad-demchuk, line 43: you already have `x` and `y` in `BaseRobot`
just add `z`

vlad-demchuk (changes requested):

PR #1882, opened 2023-06-12
Robot Factory

levchenko-dmytro:


src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
...
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    position,
```
volodymyr-soltys97, line 64: Rename 'position' parameter to 'coords' for consistency with other constructors.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1884, opened 2023-06-13
solution

HannieA:


src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion) {
```
Anastasiia-Svintsova, line 8:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion
  ) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
```
Anastasiia-Svintsova, line 49: You don't need to reassign 'this.coords' in the 'FlyingRobot' constructor, as it's already assigned in the 'BaseRobot' constructor. Instead, you can just update the 'z' property.

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return 'Robot: ' + this.name + ', Chip version: ' + this.chipVersion
+    + ', Weight: ' + this.weight;
```
Anastasiia-Svintsova, line 37: It's better to use [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) instead.

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion) {
```
Anastasiia-Svintsova, line 46: Same about the code style

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+    maxLoadWeight,
+    currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
```
Anastasiia-Svintsova, line 81: Use logical operators 
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name,
```
Anastasiia-Svintsova, line 66:
```suggestion
  constructor(
    name,
```

Anastasiia-Svintsova (changes requested):

src/robotFactory.js
HannieA: I tried, but because of limitation in 80 symbols it's needed to move a part of string to an
Screenshot 2023-06-14 at 10 52 32 AM
other line. Somehow it added indentation before new line and check failed

qoa11a (approved):

PR #1887, opened 2023-06-18
solution

vshepel:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
 
-class FlyingRobot {}
+    this.name = name;
+    this.weight = weight;
 
-class DeliveryDrone {}
+    this.coords = {
+      x: x,
+      y: y,
+    };
```
Anastasiia-Svintsova, line 13: You can use a shorthand syntax if the property name and the variable name from which you want to assign the value are the same.
```suggestion
    this.coords = {
      x,
      y,
    };
```

Anastasiia-Svintsova (approved):

PR #1897, opened 2023-06-22
solve the task

eugene-kulish:


src/robotFactory.js
```diff
@@ -1,10 +1,119 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    {
+      x = 0,
```
mykhalenych, line 8: It is better to use a default parameter for the options object to avoid potential errors. E.g. options = {}

src/robotFactory.js
```diff
@@ -1,10 +1,119 @@
...
+      y = 0,
+    },
+    chipVersion
+  ) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x,
```
mykhalenych, line 17:
```suggestion
      x: coords.x || 0,
```

src/robotFactory.js
```diff
@@ -1,10 +1,119 @@
...
+    return `Robot: ${this.name}, \
+Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(
+    name,
+    weight,
+    {
```
mykhalenych, line 49: check all constructors

PR #1902, opened 2023-06-22
robotFactory Solution

ZupaFly:


src/robotFactory.js
```diff
@@ -1,10 +1,125 @@
...
+    super.goBack(step);
+  }
+
+  goRight(step = 1) {
+    super.goRight(step);
+  }
+
+  goLeft(step = 1) {
+    super.goLeft(step);
+  }
```
Anastasiia-Svintsova, line 63: You can remove these redundant methods in the FlyingRobot class since it's already inherited from the BaseRobot class.

src/robotFactory.js
```diff
@@ -1,10 +1,125 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+
+  getInfo() {
+    super.getInfo();
+  }
```
Anastasiia-Svintsova, line 75: Same with this method

src/robotFactory.js
```diff
@@ -1,10 +1,125 @@
...
+}
+
+class DeliveryDrone extends FlyingRobot {
```
Anastasiia-Svintsova, line 78: Same comments go to the DeliveryDrone class

src/robotFactory.js
```diff
@@ -1,10 +1,125 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
+    this.coords = coords;
 
-class FlyingRobot {}
+    if (!this.coords.hasOwnProperty('x')) {
+      this.coords.x = 0;
+    }
```
Anastasiia-Svintsova, line 12: You can simplify this code by removing these if statements 
```
this.coords = {
  x: coords.x || 0,
  y: coords.y || 0,
};

Anastasiia-Svintsova (changes requested):

witflash (approved):
Well done 👍🏻

PR #1906, opened 2023-06-24
add task soulution

kbekher:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x: (coords.x) ? coords.x : 0,
```
TarasHoliuk, line 10:
```suggestion
      x: coords.x ?? 0,
```
very useful operator (perfect in this case): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
```
TarasHoliuk, line 4:
```suggestion
  constructor(name, weight, coords = {}, chipVersion) {
```
if coords are not provided at all an error will be thrown when you try to read `coords.x` or `coords.y` below

TarasHoliuk (changes requested):

src/robotFactory.js
kbekher: Thank you! I didn't know about that. It's really useful
But linter shows an error when i try to us it, therefore I'll keep my code as it is

VitaliyBondarenko1982 (approved):

PR #1910, opened 2023-06-25
add solution

akozlovska:


src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = coords;
```
OleziO, line 9: Remove this line.

OleziO (changes requested):

src/robotFactory.js
akozlovska: If the line is removed, all the test fail and coords becomes undefined
Снимок экрана 2023-06-26 в 23 29 17

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    this.coords.x = coords.x || 0;
+    this.coords.y = coords.y || 0;
```
VitaliyBondarenko1982, line 11: without this.coords will be failed tests here, because this.coords will be undefined and will not possible read field x or y in undefined.
better directly set coords with fields x, y
```suggestion
   this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
   };
```

VitaliyBondarenko1982 (changes requested):

OleziO (approved):

PR #1911, opened 2023-06-26
task solution

Bonum-taurus:


src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      z: 0,
+      ...coords,
```
volodymyr-soltys97, line 42: Avoid using spread operator to merge z coordinate. Instead, use 'coords.z' with a default value.

src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x: 0, y: 0, ...coords,
+    };
```
volodymyr-soltys97, line 11: Avoid using spread operator to merge x and y coordinates. Instead, use 'coords.x' and 'coords.y' with default values.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

anastasiiavorobiova (changes requested):
Awesome

anastasiiavorobiova (approved):

PR #1923, opened 2023-06-26
Add task solution

vladmakohon:


src/robotFactory.js
```diff
@@ -1,10 +1,114 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    { x = 0, y = 0 } = {},
```
tetiana-ivanchenko, line 7:
```suggestion
    { x = 0, y = 0 },
```

src/robotFactory.js
```diff
@@ -1,10 +1,114 @@
...
+      this.coords.z -= step;
+    }
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    { x = 0, y = 0, z = 0 } = {},
```
tetiana-ivanchenko, line 83:
```suggestion
    coords = { x: 0, y: 0, z: 0 },
```

src/robotFactory.js
```diff
@@ -1,10 +1,114 @@
...
+    currentLoad = null,
+  ) {
+    super(
+      name,
+      weight,
+      {
+        x,
+        y,
+        z,
+      },
```
tetiana-ivanchenko, line 95:
```suggestion
    super(
      name,
      weight,
      coords,
```

tetiana-ivanchenko (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,110 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(
+    name,
+    weight,
+    { x = 0, y = 0 },
```
qoa11a, line 7:
```suggestion
    coords = { x = 0, y = 0 },
```
You may declare it here.

src/robotFactory.js
```diff
@@ -1,10 +1,110 @@
...
+    chipVersion,
+  ) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x,
+      y,
+    };
```
qoa11a, line 16: Then you don't need this.

qoa11a (approved):

src/robotFactory.js
vladmakohon: @qoa11a  It seems like I may not.

Screenshot 2023-06-27 145636

And if I use a `:` then I won't be able to get default values if an empty object `{}` is passed as an argument.

tetiana-ivanchenko (approved):

PR #1939, opened 2023-06-26
Solution

OlhaSheliakina:


src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = 0, chipVersion) {
```
tetiana-ivanchenko, line 4: you can set object as default, but 0 is redundant

tetiana-ivanchenko (approved):

PR #1963, opened 2023-07-02
Solution

ArtemBehter:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+  }
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return `Robot: ${this.name}, Chip version: ${this.chipVersion}, `
```
Radoslaw-Czerniawski, line 29: You should avoid string concatenations as it might lead to some strange behaviour in js, in this case I would prefer you to disabled the max-line eslint rule as it unnecessarily complicates things
```js
// eslint-disable-next-line max-len
```

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, { x = 0, y = 0, z = 0 }, chipVersion) {
```
Radoslaw-Czerniawski, line 35: Same here with initialising

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    const coords = {
+      x: x,
+      y: y,
+    };
+
+    super(name, weight, coords, chipVersion);
```
Radoslaw-Czerniawski, line 41: You are either way rewriting these variables so might as well just:
```js
super(name, weight, { x, y }, chipVersion);
```

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    name, weight, { x = 0, y = 0, z = 0 }, chipVersion,
+    maxLoadWeight, currentLoad = null
+  ) {
+    const coords = {
+      x: x,
+      y: y,
+      z: z,
+    };
+
+    super(name, weight, coords, chipVersion);
```
Radoslaw-Czerniawski, line 66: Same here:
```js
super(name, weight, { x, y, z }, chipVersion);
```

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name, weight, { x = 0, y = 0, z = 0 }, chipVersion,
```
Radoslaw-Czerniawski, line 57:
```js
name, weight, { x = 0, y = 0, z = 0 } = {}, chipVersion,
```

Radoslaw-Czerniawski (changes requested):
Very good pr, some small changes needed

PR #1975, opened 2023-07-11
Solution

katiailchenko:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
```
ivadimko, line 4: Default parameters should be last in the function parameter list. Move 'coords = {}' as the last parameter in the constructor.

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+  goUp(step = 1) {
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
```
ivadimko, line 58: Consider adding methods to check if the drone is overloaded or if there is no cargo to unload in 'DeliveryDrone'.

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
```
ivadimko, line 72: You should check if the cargo is not null before assigning it to 'currentLoad'.

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    }
+  }
+
+  unhookLoad() {
```
ivadimko, line 78: You might want to return the unloaded cargo from 'unhookLoad' method.

ivadimko (approved):

PR #1978, opened 2023-07-13
Solution

Luk2asz:


src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x,
+      y: coords.y,
+    };
```
DorotaLeniecDev, line 11: You could simplify:
```js
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    let resultUnder80 = `Robot: ${this.name}, `;
```
DorotaLeniecDev, line 40: Why this varaiable is called ``resultUnder80``? Maybe it could be named more descriptively?

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+    resultUnder80 += `Weight: ${this.weight}`;
+
+    return resultUnder80;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z;
```
DorotaLeniecDev, line 52: Make use of ```||```

```js
```

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    }
+  }
```
DorotaLeniecDev, line 80:
```suggestion
    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
      this.currentLoad = cargo;
    }
```

DorotaLeniecDev (changes requested):
Few changes needed

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    }
+  }
+
+  unhookLoad() {
```
darokrk, line 71: The 'unhookLoad' method does not check if there is a current load before setting 'this.currentLoad' to null. It can return a message or throw an error when there is no current load to unhook. Consider to handle it

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords,
+    chipVersion, maxLoadWeight, currentLoad = null) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
+    this.currentLoad = currentLoad;
+  }
+
+  hookLoad(cargo) {
+    if (!this.currentLoad && cargo.weight <= this.maxLoadWeight) {
```
darokrk, line 66: The 'hookLoad' method does not handle the situation when the cargo's weight exceeds the maximum load weight. It can throw an error or return a message in this case. Consider to handle it

darokrk (approved):
Looks nice 👍🏽  left 2 mentions to consider for improvement, you can fix them if you want to, but are not necessary 😃

PR #1980, opened 2023-07-13
Solution

ArtemVelychko:


src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
+    if (coords.x === undefined) {
+      coords.x = 0;
+    }
```
polosanya, line 7:
```suggestion
    this.coords = {
      x: coords.x || 0,
      ......
    };
```
You can add all coords at once 
image

src/robotFactory.js
```diff
@@ -1,10 +1,99 @@
...
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
+    this.chipVersion = chipVersion;
+  }
+
+  goForward(step = 1) {
+    this.coords.y += step;
+
+    return this.coords.y;
```
polosanya, line 22:
```suggestion
  goForward(step = 1) {
    this.coords.y += step;
```
Movement methods shouldn't return anything

polosanya (changes requested):
Good job! Let's improve your code a bit

PR #1983, opened 2023-07-15
Solution

yuliia-nudyk:


src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  getInfo() {
+    return 'Robot: ' + this.name + ', '
```
lerastarynets, line 32: Consider using template literals (backticks) for string concatenation instead of mixing '+' operator and backticks. It will not only make the code cleaner but also easier to read and modify.

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

src/robotFactory.js
yuliia-nudyk: I usually use backtick, but it was too large string for linter...

PR #1985, opened 2023-07-18
Solution

Hladkyi03:


src/robotFactory.js
```diff
@@ -1,6 +1,23 @@
...
+    this.chipVersion = chipVersion;
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
+  }
+
+  getInfo() {
+    // eslint-disable-next-line max-len
```
loralevitska, line 17: Try to fix eslint rule and remove comment, for example destructure values from `this` and make variables

loralevitska (changes requested):
A little more and it will be completed
Next time make PR to master branch to avoid conflicts in PR
image

etojeDenys (approved):
lgtm

PR #1996, opened 2023-07-24
Robot Factory task solution

AnVBondar:


src/robotFactory.js
```diff
@@ -1,10 +1,135 @@
...
+    name,
+    weight,
+    { x = 0, y = 0, z = 0 } = {},
+    chipVersion
+  ) {
+    super(name, weight, true, chipVersion);
+
+    this.coords = {
+      x,
+      y,
```
vlad-demchuk, line 69: you should inherit x, y and just add z here

src/robotFactory.js
```diff
@@ -1,10 +1,135 @@
...
+    maxLoadWeight,
+    currentLoad = null,
+  ) {
+    super(name, weight, true, chipVersion);
+
+    this.coords = {
+      x,
+      y,
+      z,
+    };
```
vlad-demchuk, line 106: just add coords to super

vlad-demchuk (changes requested):

src/robotFactory.js
```diff
@@ -1,10 +1,131 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(
+    name,
+    weight,
+    { x = 0, y = 0, z = 0 } = {},
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null,
+  ) {
+    super(name, weight, true, chipVersion);
```
volodymyr-soltys97, line 102: You are not passing the correct parameters to the 'super' function. It should receive name, weight, coords, and chipVersion, but you are passing 'true' instead of the coordinates.

src/robotFactory.js
```diff
@@ -1,10 +1,131 @@
...
+  hookLoad(cargo) {
+    if (this.currentLoad !== null) {
+      return;
+    };
+
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    } else {
+      return;
+    };
```
volodymyr-soltys97, line 122: The semicolon after the function declaration is unnecessary. You can remove it.

src/robotFactory.js
```diff
@@ -1,10 +1,131 @@
...
+  };
+
+  unhookLoad() {
+    this.currentLoad = null;
+  };
```
volodymyr-soltys97, line 127: The semicolon after the function declaration is unnecessary. You can remove it.

src/robotFactory.js
```diff
@@ -1,10 +1,131 @@
...
+
+  hookLoad(cargo) {
+    if (this.currentLoad !== null) {
+      return;
+    };
+
+    if (this.currentLoad === null && cargo.weight <= this.maxLoadWeight) {
+      this.currentLoad = cargo;
+    } else {
+      return;
```
volodymyr-soltys97, line 121: There's no need to return anything in the 'hookLoad' function. You can remove the 'else' block.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

volodymyr-soltys97 (approved):
Good job 👍

PR #1997, opened 2023-07-25
Solution

RomanenkoStud:


src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+      + `Chip version: ${this.chipVersion}, `
+      + `Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const defaultCoords = {
```
volodymyr-soltys97, line 47: The defaultCoords object in the FlyingRobot subclass is redundant. The parent class, BaseRobot, has already assigned the default values to the coordinates.

src/robotFactory.js
```diff
@@ -1,10 +1,88 @@
...
+      x: 0,
+      y: 0,
+      z: 0,
+    };
+
+    this.coords = {
```
volodymyr-soltys97, line 53: The coords assignment in FlyingRobot is unnecessary. The BaseRobot class, from which FlyingRobot is inheriting, already handles the assignment of coordinates.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

volodymyr-soltys97 (approved):
Nice work 👍

PR #1999, opened 2023-07-26
Solution

BeeGooDee:


src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
+    coords,
+    chipVersion
+  ) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = coords;
+    coords.x = coords.x || 0;
+    coords.y = coords.y || 0;
+    this.chipVersion = chipVersion;
+  }
```
etojeDenys, line 16: try to rewrite using [Typescript Constructor Shorthand](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd)

src/robotFactory.js
```diff
@@ -1,10 +1,92 @@
...
 
-class FlyingRobot {}
+  goForward(step = 1) {
+    this.coords.y += step;
+  }
```
etojeDenys, line 20: let's add the check if the step is not negative

etojeDenys (changes requested):

OleziO (approved):

PR #2010, opened 2023-08-03
Robot Factory

KseniaSkr:


src/robotFactory.js
```diff
@@ -1,10 +1,65 @@
...
+    this.coords.y += step;
+  }
+  goBack(step = 1) {
+    this.coords.y -= step;
+  }
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
```
etojeDenys, line 23: add the check if step is not negative

etojeDenys (changes requested):

IvanFesenko (approved):
Well done

PR #2015, opened 2023-08-06
Solution

Patryk-Filipiak:


szymonpachucki (commented):
looks good to me

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+
+class DeliveryDrone extends FlyingRobot {
+  constructor() {
+    super(...arguments);
+
+    Object.assign(this, {
+      maxLoadWeight: arguments[4],
+      currentLoad: arguments[5] || null,
+    });
+  }
```
asakevych, line 67: I think it would be more readable if you write every argument name instead of using `arguments[4]`
```suggestion
   constructor(name, weight, coords, chipVersion, maxLoadWeight, currentLoad) {
    super(name, weight, coords, chipVersion);
    this.maxLoadWeight = maxLoadWeight;
    this.currentLoad = currentLoad || null;
  }
```

asakevych (changes requested):
Good job!! just one small thing to change 😉

src/robotFactory.js
Patryk-Filipiak: Changed in that class and in FlyingRobot too :)

asakevych (approved):
Great job!! 🚀 🔥 👏

PR #2016, opened 2023-08-06
Develop

szymonpachucki:


src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+  }, chipVersion) {
+    Object.assign(this, {
+      name,
+      weight,
+      coords: {
+        x,
+        y,
+      },
+      chipVersion,
+    });
```
choeqq, line 16: I think it would be better to assign each property separately:
```suggestion
    this.name = name;
    this.weight = weight 
    etc...
```

choeqq (changes requested):

choeqq (approved):

PR #2021, opened 2023-08-07
robotFactory task solution added

Serhii-Dr:


src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    const { z = 0 } = coords;
+
+    this.coords = Object.assign(this.coords, { z });
```
etojeDenys, line 43:
```suggestion
    const { z = 0 } = coords;

    this.coords.z = z
```

src/robotFactory.js
```diff
@@ -1,10 +1,74 @@
...
+    this.coords.y += step;
+  }
+  goBack(step = 1) {
+    this.coords.y -= step;
+  }
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
+  goLeft(step = 1) {
+    this.coords.x -= step;
```
etojeDenys, line 26: add some checks if step is not negative

etojeDenys (changes requested):

Viktor-Kost (approved):

PR #2027, opened 2023-08-09
solutionRobotFactory

anastasiia-burkovska:


src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, position, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.coords = {};
+    this.coords.x = position.x || 0;
+    this.coords.y = position.y || 0;
```
etojeDenys, line 9:
```suggestion
    this.coords = {
      x: position.x || 0;
      y: position.y || 0;
    }
```

src/robotFactory.js
```diff
@@ -1,10 +1,76 @@
...
+    this.coords.y -= step;
+  }
+
+  goLeft(step = 1) {
+    this.coords.x -= step;
+  }
+
+  goRight(step = 1) {
+    this.coords.x += step;
+  }
```
etojeDenys, line 33: let's add the check if the step is not negative here

etojeDenys (changes requested):

VitaliyBodnar (approved):
Good job!

PR #2030, opened 2023-08-10
Solution

klaudia-klis:


src/robotFactory.js
```diff
@@ -1,10 +1,72 @@
...
+    this.coords.z += step;
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords = 0,
```
mvjl000, line 53: I think the default value of coords should be `0`

mvjl000 (changes requested):
Just a minor change needed. Besides of that you did great! 🚀

src/robotFactory.js
klaudia-klis: I thought that because it extends BaseRobot class, it will take the default value from this code: 
`    this.coords = {
       x: coords.x || 0,
       y: coords.y || 0,
`
But I added 0 on line 53 as you suggested.

choeqq (approved):

PR #2033, opened 2023-08-10
Solution

kabakTech:


src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
```
choeqq, line 4: I think it would be better to have default coords object, you know how it will look like 😄 
```suggestion
  constructor(name, weight, coords = {z: 0, y:0}, chipVersion) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,77 @@
...
+      `Chip version: ${this.chipVersion},`,
+      `Weight: ${this.weight}`,
+    ].join(' ');
+
+    return info;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords = {}, chipVersion) {
```
choeqq, line 43: you can also have default object in here, please fill it up with values

choeqq (changes requested):

choeqq (changes requested):
Waiting for any updates @jakubMikolajczykTech 👀

choeqq (changes requested):
Waiting for any updates @jakubMikolajczykTech 👀

PR #2036, opened 2023-08-10
solution

PawelP2001:


src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, {
+    x = 0, y = 0,
+  }, chipVersion) {
```
choeqq, line 6: you should use coords object to get x and y
```suggestion
  constructor(name, weight, coords = {x: 0, y: 0}, chipVersion) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,87 @@
...
+
+  getInfo() {
+    return `Robot: ${this.name}, `
+    + `Chip version: ${this.chipVersion}, Weight: ${this.weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, {
+    x = 0, y = 0, z = 0,
```
choeqq, line 40: same with this object ⬆️

choeqq (changes requested):

PR #2042, opened 2023-08-10
develop

MateuszSlezak:


src/robotFactory.js
```diff
@@ -1,10 +1,75 @@
...
+    const chip = `Chip version: ${this.chipVersion}`;
+
+    return (`${name}, ${chip}, Weight: ${this.weight}`);
+  }
+};
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+    this.coords = coords;
```
choeqq, line 43: you should either assing the whole object to `coords` or each property one by one, you have to choose which one you're using, also please do not forget about assigning default values for other coords

choeqq (changes requested):

choeqq (approved):

PR #2063, opened 2023-08-15
add task solution

Koliras:


src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
 'use strict';
 
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, { x = 0, y = 0 }, chipVersion) {
```
33Orange, line 4: You simply assign a new object with all the fields. If you are given an empty object as a class argument, it will override your default object.

```suggestion
  constructor(name, weight, coords, chipVersion) {
```

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x,
+      y,
+    };
```
33Orange, line 12:
```suggestion
    this.coords = {
      x: coords.x || 0,
      y: coords.y || 0,
    };
```

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, { x = 0, y = 0, z = 0 }, chipVersion) {
+    super(name, weight, {
+      x, y,
+    }, chipVersion);
+
+    this.coords.z = z;
+  }
```
33Orange, line 45: Same here

src/robotFactory.js
```diff
@@ -1,10 +1,84 @@
...
+    name,
+    weight,
+    { x = 0, y = 0, z = 0 },
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null
+  ) {
+    super(name, weight, {
+      x, y, z,
+    }, chipVersion);
```
33Orange, line 67:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
    super(name, weight, coords, chipVersion);
```

33Orange (changes requested):
Almost done

tiserett (approved):

PR #2075, opened 2023-08-17
add solution

OlhaMomot:


src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
-class BaseRobot {}
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
tiserett, line 10: I'd like you to create a variable for default coords

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+      z: coords.z || 0,
+    };
```
tiserett, line 45: you can just set this.coords.z instead of reseting whole object

src/robotFactory.js
```diff
@@ -1,10 +1,78 @@
...
+  }
+
+  goDown(step = 1) {
+    this.coords.z -= step;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight = null, currentLoad = null) {
```
tiserett, line 59: is there a better way to format this line?

tiserett (changes requested):

PR #2083, opened 2023-08-19
solution

Valentyn-Bilyk:


src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
+    };
 
-class DeliveryDrone {}
+    this.chipVersion = chipVersion;
```
Moroz-Dmytro, line 18: better to place it on line 12, after `this.weight = weight;`

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+    coords,
+    chipVersion
+  ) {
+    this.name = name;
+    this.weight = weight;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
Moroz-Dmytro, line 15: Would be a good idea to create constant for default values

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+
+class FlyingRobot extends BaseRobot {
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.coords.z = coords.z || 0;
```
Moroz-Dmytro, line 52: The same about default value

Moroz-Dmytro, line 52:
```suggestion
    super(name, weight, coords, chipVersion);
    
    this.coords.z = coords.z || 0;
```

src/robotFactory.js
```diff
@@ -1,10 +1,89 @@
...
+class DeliveryDrone extends FlyingRobot {
+  constructor(name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null,
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
```
Moroz-Dmytro, line 73:
```suggestion
    super(name, weight, coords, chipVersion);
    
    this.maxLoadWeight = maxLoadWeight;
```

Moroz-Dmytro (changes requested):
GJ 👍 
Some comments below

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
 'use strict';
 
-class BaseRobot {}
+const DEFAULT_X = 0;
+const DEFAULT_Y = 0;
+const DEFAULT_Z = 0;
```
Moroz-Dmytro, line 5:
```suggestion
const DEFAULT_COORDS = {
  x: 0,
  y: 0,
  z: 0,
}
```

Moroz-Dmytro: But it would be better

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+    this.chipVersion = chipVersion;
 
-class DeliveryDrone {}
+    this.coords = {
+      x: coords.x || DEFAULT_X,
+      y: coords.y || DEFAULT_Y,
+    };
+  }
+
+  goForward(step = 1) {
```
Moroz-Dmytro, line 24: And for this value we also could create a constant

Moroz-Dmytro (approved):
Good work 👍 
But check comments

PR #2087, opened 2023-08-20
Develop

edgergard:


src/robotFactory.js
```diff
@@ -1,10 +1,90 @@
...
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
 
-class DeliveryDrone {}
+    this.coords = {
+      x: coords.x || ROBOT_BASE_COORD,
+      y: coords.y || ROBOT_BASE_COORD,
+    };
+    this.chipVersion = chipVersion;
```
Moroz-Dmytro, line 15: Better to place it on line 10, after `this.weight = weight;`

src/robotFactory.js
```diff
@@ -1,10 +1,90 @@
...
+  goDown(step = BASE_STEP) {
+    this.coords.z -= step;
+
+    return this;
+  }
+}
+
+class DeliveryDrone extends FlyingRobot {
+  constructor(name, weight, coords, chipVersion,
+    maxLoadWeight, currentLoad = null) {
```
Moroz-Dmytro, line 71:
```suggestion
  constructor(
    name,
    weight,
    coords,
    chipVersion,
    maxLoadWeight,
    currentLoad = null
  ) {
```

Moroz-Dmytro: Remember about good formatting

src/robotFactory.js
```diff
@@ -1,10 +1,90 @@
 'use strict';
 
-class BaseRobot {}
+const ROBOT_BASE_COORD = 0;
```
Moroz-Dmytro, line 3:
```suggestion
const ROBOT_BASE_COORDS = {
  x: 0,
  y: 0,
  z: 0,
};
```

Moroz-Dmytro: What if base coords can be different?

Moroz-Dmytro (changes requested):
GJ 👍 
CHeck comments below

roman-mirzoian (approved):
Well done 👍

PR #2089, opened 2023-08-20
Solution

yuliiaant:


src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
Moroz-Dmytro, line 11: Would be a good idea to create a constant with default coords and use it here

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    super(name, weight, coords, chipVersion);
+
+    this.coords = {
+      ...coords,
+      z: coords.z || 0,
+    };
```
Moroz-Dmytro, line 45:
```suggestion
    this.coords.z = coords.z || 0
```

src/robotFactory.js
```diff
@@ -1,10 +1,83 @@
...
+  constructor(
+    name,
+    weight,
+    coords,
+    chipVersion,
+    maxLoadWeight,
+    currentLoad = null,
+  ) {
+    super(name, weight, coords, chipVersion);
+    this.maxLoadWeight = maxLoadWeight;
```
Moroz-Dmytro, line 67:
```suggestion
    super(name, weight, coords, chipVersion);
    
    this.maxLoadWeight = maxLoadWeight;
```

src/robotFactory.js
Moroz-Dmytro: But the same comment about default value

Moroz-Dmytro (changes requested):
GJ 👍 
Small comments

tiserett (approved):

PR #2090, opened 2023-08-21
Solution

valet2102:


src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
 'use strict';
 
-class BaseRobot {}
+const ROBOT_BASE_COORD = 0;
```
roman-mirzoian, line 3: What if we want to change one of the base coordinates?
```suggestion
const ROBOT_BASE_COORDS = {
  x: 0,
  y: 0,
  z: 0
};
```

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+    this.coords = {
+      x: position.x || ROBOT_BASE_COORD,
+      y: position.y || ROBOT_BASE_COORD,
+    };
+  }
+
+  goForward(step = ROBOT_BASE_STEP) {
+    this.coords.y += step;
+
+    return this;
```
roman-mirzoian, line 21: In this task, we don't need to return anything in the class methods.

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+    const { name, chipVersion, weight } = this;
+
+    return `Robot: ${name}, Chip version: ${chipVersion}, Weight: ${weight}`;
+  }
+}
+
+class FlyingRobot extends BaseRobot {
+  constructor(name, weight, position, chipVerion) {
+    super(name, weight, position, chipVerion);
+    this.coords.z = position.z || ROBOT_BASE_COORD;
```
roman-mirzoian, line 52:
```suggestion

    this.coords.z = position.z || ROBOT_BASE_COORD;
```

src/robotFactory.js
```diff
@@ -1,10 +1,94 @@
...
+  constructor(
+    name,
+    weight,
+    position,
+    chipVerion,
+    maxLoadWeight,
+    currentLoad = null,
+  ) {
+    super(name, weight, position, chipVerion);
+    this.maxLoadWeight = maxLoadWeight;
```
roman-mirzoian, line 78:
```suggestion

    this.maxLoadWeight = maxLoadWeight;
```

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

PR #2092, opened 2023-08-21
add task solution

voievodik:


src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+class BaseRobot {
+  constructor(name, weight, coords, chipVersion) {
+    this.name = name;
+    this.weight = weight;
+    this.chipVersion = chipVersion;
 
-class FlyingRobot {}
+    this.coords = {
+      x: coords.x || 0,
+      y: coords.y || 0,
```
Moroz-Dmytro, line 11: Would be nice to create constants for default coords

src/robotFactory.js
```diff
@@ -1,10 +1,81 @@
...
+    };
+  }
 
-class DeliveryDrone {}
+  goForward(step = 1) {
```
Moroz-Dmytro, line 15: Would be nice to create constant for this magic value

Moroz-Dmytro (approved):
GJ 👍 
Small comments for improvement

